PR #8, opened 2020-12-05
Added a new message, and a check for url messages

PrestonEn:
likely a naive format for checking for URL content in messages, consider a refactor?

### DIFF
--- a/duckbot/days.py
+++ b/duckbot/days.py
@@ -1,6 +1,7 @@
 
 import datetime
 import random
+import validators
 
 def announce_day_of_week():
     days = {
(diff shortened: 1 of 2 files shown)


duckbot/days.py
```diff
@@ -45,7 +48,11 @@ def announce_day_of_week():
     now = datetime.datetime.now()
     if str(now.hour) == "7":
         day = datetime.datetime.today().strftime('%A')
-        return random.choice(templates).format(random.choice(days[day]))
+        message = random.choice(days[day])
+
+        if validators.url(message):
+            return f"A video to start your {day}: {message}"
+        return random.choice(templates).format()
```
twentylemon, line 55: Should be `format(message)`

PR #16, opened 2020-12-11
DuckBot now announces its presence

Chippers255:


twentylemon:
I'm against this change. I don't really want noise for the sake of noise.

Chippers255:
I figured until the bot is hosted it would be good to know when it is online. What about pushing this message to the DuckButt channel?

twentylemon:
Posting to #duckbutt is fine I suppose. At some point though, the bot will always be online; announcing it's presence is moot. I'd rather avoid useless noise for no real reason.

Chippers255:
Fair enough, I will revert.

jameshughes89:
Fucking engineers

twentylemon:
Rude.

PR #29, opened 2020-12-18
Send daily greeting at 7am eastern time

### LINKED ISSUE #27: daily greeting is using 7am UTC instead of 7am EST

twentylemon:
Fixes #27


twentylemon:
Ah, F. I forgot why I didn't open this earlier. Apparently python modules suck. Need to fix locally, I'll re-open later.

twentylemon:
> Ah, F. I forgot why I didn't open this earlier. Apparently python modules suck. Need to fix locally, I'll re-open later.

Fixed. Jamming the src root into the system path during tests allows us to use the same import style in src and tsts.

Chippers255 (approved):

PR #68, opened 2021-02-23
Removing self hosted runner for security reasons

Chippers255:


twentylemon (approved):
Let the ducking begin.

PR #90, opened 2021-03-13
Add docker-test script, which builds/tests duckbot in docker

twentylemon:
The point here is to aid with the deployment side of things. We can run this script (well, the `Dockerfile` at least) before the deployment to ensure duckbot will actually work, and we won't bring it down by performing the deployment.


Chippers255:
Update the branch brother

Chippers255 (approved):

PR #102, opened 2021-03-24
add !start and !stop commands for duckbot to stream "music"

twentylemon:
This adds `!start` and `!stop` commands which will tell duckbot to enter _Hangout 1_ and start streaming music, and to leave after that. Playing music requires `ffmpeg`, so I added that to the install for the docker image and other scripts to run the bot. It doesn't need to be explicitly added to the github actions workflow since it's already in the `install` script.

I also added a `resources` directory/module. The songs are currently just pulled from there.

On a side note, holy commits batman!


twentylemon:
Answering @Chippers255  from the discord:
> what is the point of the empty test_resource file in the resources folder? I assume for test coverage.

It's for code coverage essentially, yes. I haven't looked too much into python resources, I just whipped up the `Resouces` cog to read them. The cog method accepts `paths*`, but we've no resources that are in subdirectories. It's a blob of nothing to ensure reading resources in subdirectories would work.

Ie, the class supports:
```python
resources.get("thing.txt")  # resources/thing.txt
resources.get("a", "b", "c.txt") # resources/a/b/c.txt
```
but the only prod resources are of style `thing.txt`.

twentylemon:
> Answering @Chippers255 from the discord:
> 
> > what is the point of the empty test_resource file in the resources folder? I assume for test coverage.
> 
> It's for code coverage essentially, yes. I haven't looked too much into python resources, I just whipped up the `Resouces` cog to read them. The cog method accepts `paths*`, but we've no resources that are in subdirectories. It's a blob of nothing to ensure reading resources in subdirectories would work.
> 
> Ie, the class supports:
> 
> ```python
> resources.get("thing.txt")  # resources/thing.txt
> resources.get("a", "b", "c.txt") # resources/a/b/c.txt
> ```
> 
> but the only prod resources are of style `thing.txt`.

TBH I forgot mocking static methods was a thing in python. Removed the test resource and instead stubbed `os.path.exists` in tests.

twentylemon:
> TBH I forgot mocking static methods was a thing in python. Removed the test resource and instead stubbed `os.path.exists` in tests.

Or, better yet, using [importlib](https://docs.python.org/3/library/importlib.html#module-importlib.resources) for resources instead of a cog.

Chippers255 (approved):

PR #104, opened 2021-03-25
add -u flag when running duckbot

twentylemon:
Iunno why but `-u` seems like it might fix things.

https://docs.python.org/3/using/cmdline.html#cmdoption-u


zambon:
LGTM

Chippers255 (approved):

PR #140, opened 2021-04-15
round rain chance and precipitation amount to whole number

twentylemon:
Floating point math sometimes gives annoying precipitation changes,

> with a 56.99999999999999% chance of 0.92mm of rain

should change to

> with a 57% chance of 1mm of rain


Chippers255:
@twentylemon you don't exactly have to using rounding or ceiling in python. The fstring can do decimal place rounding. For example:
```
chance = 0.5677777777777777777777777777777777777
print(f"The chance of weather is {chance:.2%}.")
print(f"The chance of weather is {chance:.2f}.")
```
would produce the following:
```
The chance of weather is 56.77%.
The chance of weather is 0.57.
```

twentylemon:
> @twentylemon you don't exactly have to using rounding or ceiling in python. The fstring can do decimal place rounding. For example:
> 
> ```
> chance = 0.5677777777777777777777777777777777777
> print(f"The chance of weather is {chance:.2%}.")
> print(f"The chance of weather is {chance:.2f}.")
> ```
> 
> would produce the following:
> 
> ```
> The chance of weather is 56.77%.
> The chance of weather is 0.57.
> ```

Meh. I tend to prefer to prefer the syntax you don't need to pull out a textbook to understand. `round` and `ceil` are much simpler to read than f-string magic.

I can change the code if you feel strongly, but I don't see any strong reason to prefer one over the other in this case.

Chippers255:
> > @twentylemon you don't exactly have to using rounding or ceiling in python. The fstring can do decimal place rounding. For example:
> > ```
> > chance = 0.5677777777777777777777777777777777777
> > print(f"The chance of weather is {chance:.2%}.")
> > print(f"The chance of weather is {chance:.2f}.")
> > ```
> > 
> > 
> > would produce the following:
> > ```
> > The chance of weather is 56.77%.
> > The chance of weather is 0.57.
> > ```
> 
> Meh. I tend to prefer to prefer the syntax you don't need to pull out a textbook to understand. `round` and `ceil` are much simpler to read than f-string magic.
> 
> I can change the code if you feel strongly, but I don't see any strong reason to prefer one over the other in this case.

In this case I am not too concerned, the reason for the string format rounding is so you don't change an underlying value. In our case it doesn't matter.

Chippers255 (approved):

PR #163, opened 2021-04-24
avoid accessing `channel.name` if it doesn't exist

twentylemon:
`channel.name` is not defined for direct messages, this is just avoiding the `AttributeNotDefined` error for the dank cog.


twentylemon:
Moved to #158

PR #175, opened 2021-05-03
delete edit-diff messages after 5min

twentylemon:
Just to clean up the chat, duckbot will delete the edit-diff messages after a bit.

I'm definitely flexible on the timing. I can see something like 30s being nice too. Let me know your thoughts.


Chippers255:
5min is good for now. Let's run it for a week or so and see.

Chippers255 (approved):

PR #193, opened 2021-05-06
detect if a message was a haiku

### LINKED ISSUE #79: Duckbot should detect when a message is a haiku
Duckbot should count syllables for messages. If the message has the correct amount syllables for a haiku it should return that message formatted as a haiku.

> Dave: It's a little weird. I think "kids these days" are more needy in general.
> DuckBot: It's a little weird.
>                I think "kids these days" are more
>                Needy in general.

twentylemon:
Fixes #79 

~~I ended up finding [Carnegie Mellon Pronouncing Dictionary](http://www.speech.cs.cmu.edu/cgi-bin/cmudict), and it also happens to be on [github](https://github.com/Alexir/CMUdict/blob/master/cmudict-0.7b). It lists pronunciation and emphasis of some 130k words. They also have a [tool](http://www.speech.cs.cmu.edu/tools/lextool.html) for trying to pronounce new words not in the dictionary, which could be an extension to what I have done here.~~

nltk has the cmudict built in, so I'm using that now instead. nltk is a huge pain to stub out when testing, I couldn't figure out a way to get it running without downloading the packages (it literally checks when you import the package, so you can't even mock it out). I eventually just added the package download to the `install` script so it'd work.

### DIFF
--- a/duckbot/cogs/messages/haiku.py
+++ b/duckbot/cogs/messages/haiku.py
@@ -1,0 +1,59 @@
+from discord.ext import commands
+import urllib
+from discord import Embed, Colour
+
+
+class Haiku(commands.Cog):
+    def __init__(self, bot):
+        self.bot = bot
+        self.syllables = {}
+
+    @commands.Cog.listener("on_ready")
+    async def build_syllable_dictionary(self):
+        data = urllib.request.urlopen(urllib.request.Request("https://raw.githubusercontent.com/Alexir/CMUdict/master/cmudict-0.7b"))
+        syllables = {}
+        for bline in data:
+            line = bline.decode("latin1").strip()
+            if line and line[0] != ";":  # ; are comments
+                tokens = line.split()
+                count = len([t for t in tokens[1:] if t[-1].isdigit()])
+                syllables[tokens[0].lower()] = count
+        self.syllables = syllables
+
+    @commands.Cog.listener("on_message")
+    async def detect_haiku(self, message):
+        if message.author == self.bot.user:
+            return
+        words = self.get_words(message)
+        i = 0
+        lines = []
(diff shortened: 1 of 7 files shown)


twentylemon:
Moving to draft. Seems nltk has the cmudict built in!
* https://github.com/nltk/nltk/blob/develop/nltk/corpus/reader/cmudict.py

I'll opt to use that instead of parsing it myself.

duckbot/cogs/messages/haiku.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Haiku(commands.Cog):
+    def __init__(self, bot):
+        self.bot = bot
+        self.syllables = {}
+
+    @commands.Cog.listener("on_ready")
+    async def build_syllable_dictionary(self):
+        nltk.download("cmudict")
```
twentylemon, line 14: Think a bit more about this download, maybe we should put this into the dockerfile for duckbot? What are your thoughts?  
We'd still need to download it for the `install` script, but it'd save this re-download every deployment.

twentylemon: Can apparently also use `setup.py` to install this kind of stuff. Maybe worth looking into later.

Chippers255: I'll dig up an old dockerfile we used to use at work, we were able to reload the dockerfile with the NLTK downloads.

twentylemon: Threw together something [on a dev branch](https://github.com/twentylemon/duckbot/pull/12/files) which uses `setuptools` to ensure the nltk data is downloaded. Think for now, we can just leave this PR as is, and revisit when I post a PR for that branch.

Chippers255 (approved):

PR #208, opened 2021-05-13
configure discord-py logging

jameshughes89:
Small change to enable logging. Code is more or less copy/pasteed from discord.py tutorial.


twentylemon:
Logging -> adds mp3 files. What?

jameshughes89:
> Logging -> adds mp3 files. What?

Those were intended to be in a different branch (tito).

Chippers255:
Respect the branch name brother.

duckbot/__main__.py
```diff
@@ -19,6 +20,15 @@
 import duckbot.util.connection_test
 
 
+# Logging setup
+# Stole right from Discord
+logger = logging.getLogger("discord")
+logger.setLevel(logging.DEBUG)
+handler = logging.FileHandler(filename="connie.log", encoding="utf-8", mode="w")
+handler.setFormatter(logging.Formatter("%(asctime)s:%(levelname)s:%(name)s: %(message)s"))
+logger.addHandler(handler)
```
twentylemon: 1. Remove comments (also, discord-py is _not_ discord)
2. I don't think we want logs stored in files without log rotation/invalidation.
3. This shouldn't run outside of main

twentylemon:
> > Logging -> adds mp3 files. What?
> 
> Those were intended to be in a different branch (tito).

They are not related to _logging_, so remove them from this PR.

duckbot/__main__.py
jameshughes89: > 1. Remove comments (also, discord-py is _not_ discord)
> 2. I don't think we want logs stored in files without log rotation/invalidation.
> 3. This shouldn't run outside of main

This looks like a nice easy way to get log rotation: https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler

twentylemon: Brother, I believe in you. Cram it into this PR.

jameshughes89:
All entirely untested.

duckbot/__main__.py
```diff
@@ -56,6 +57,17 @@ def intents() -> Intents:
     return intent
 
 
+def logger_setup():
+    logger = logging.getLogger("discord")
+    logger.setLevel(logging.DEBUG)
```
twentylemon: Given discord-py uses [gateway](https://discord.com/developers/docs/topics/gateway) events, debug is likely to be littered with a lot of heartbeat responses. I don't know if discord-py logs those at debug or info, this may even be better bumping up to warning. Something that will need to be tested. We'll get you a dev token at some point so you can dig a little deeper.

duckbot/__main__.py
```diff
@@ -56,6 +57,17 @@ def intents() -> Intents:
...
+
+    # The choice of maxBytes=256000=250kB is arbitrary
```
twentylemon: Please remove comment. Inline comments such as this are meant to clarify complex logic or decisions. When I come back to this piece of code in two years, I'm not going to ask why the log file was capped to 250kb, so this comment is just noise I have to filter out.

twentylemon (commented):
Looks much better! Almost like a real dev. 🚀 

I'd like to know the content of those logs before approval, we'll have to get you an api token.

duckbot/__main__.py
```diff
@@ -56,6 +57,17 @@ def intents() -> Intents:
...
+    handler = logging.RotatingFileHandler(filename="./logs/duck.log", mode="a", maxBytes=256000, backupCount=10)
```
twentylemon: Seems this is `logging.handlers.RotatingFileHandler`, see [sanity test](https://github.com/Chippers255/duckbot/runs/2578419360#step:3:950) and [docs](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler)

twentylemon (changes requested):

duckbot/__main__.py
jameshughes89:
```
2021-05-09 16:10:26,342:DEBUG:discord.http: GET https://discord.com/api/v7/gateway with None has returned 200
2021-05-09 16:10:26,343:DEBUG:discord.http: GET https://discord.com/api/v7/gateway has received {'url': 'wss://gateway.discord.gg'}
2021-05-09 16:10:26,551:DEBUG:discord.gateway: Created websocket connected to wss://gateway.discord.gg?encoding=json&v=6&compress=zlib-stream
2021-05-09 16:10:26,551:DEBUG:discord.client: Dispatching event socket_raw_receive
2021-05-09 16:10:26,551:DEBUG:discord.gateway: For Shard ID None: WebSocket Event: {'t': None, 's': None, 'op': 10, 'd': {'heartbeat_interval': 41250, '_trace': ['["gateway-prd-main-vf9c",{"micros":0.0}]']}}
2021-05-09 16:10:26,552:DEBUG:discord.client: Dispatching event socket_response
2021-05-09 16:10:26,552:DEBUG:discord.client: Dispatching event socket_raw_send
2021-05-09 16:10:26,552:DEBUG:discord.client: Dispatching event socket_raw_send
2021-05-09 16:10:26,553:INFO:discord.gateway: Shard ID None has sent the IDENTIFY payload.
2021-05-09 16:10:26,633:DEBUG:discord.client: Dispatching event socket_raw_receive
2021-05-09 16:10:26,633:DEBUG:discord.gateway: For Shard ID None: WebSocket Event: {'t': None, 's': None, 'op': 11, 'd': None}
```


Example debug level output above. Literally says `heartbeat` --- good call.

twentylemon: debug is also logging every api call, I don't think we really need request level logging. INFO is probably fine for now, we can always revisit in the future.

jameshughes89: Agreed. INFO also seems overkill, but not too verbose.

duckbot/__main__.py
jameshughes89: What would a reasonable log file size cap be? If level is changed to INFO, logs should be much smaller.

twentylemon: More of a @Chippers255 question. It's his hardware. ~4mb though (10 x 250kb), seems fine as is. That probably ends up being quite a long time range stored in logs, given we don't really have a huge amount of interactions.

duckbot/__main__.py
```diff
@@ -56,6 +58,16 @@ def intents() -> Intents:
     return intent
 
 
+def logger_setup():
+    logger = logging.getLogger("discord")
+    logger.setLevel(logging.INFO)
+
+    handler = logging.handlers.RotatingFileHandler(filename="./logs/duck.log", mode="a", maxBytes=256000, backupCount=10)
```
twentylemon: Checks still failing, guessing it doesn't like the `logs` directory. I imagine the append writing mode would create the log files, but not intermediary directories.

The only stuff in `.` is `duckbot`, `resources` and `requirements.txt`, so we're probably safe to simply use `duck.log` -- or creating the intermediary directories.

Further, thinking more about this, we'll need to add the log files to gitignore, in case people run duckbot from outside of docker. Either the logs directory (if it's used), or `*.log*` should be added -- or both.

twentylemon (changes requested):

duckbot/__main__.py
jameshughes89: I like the idea of creating the logs dir with .gitignore.

jameshughes89:
`makedirs()` will recursively make directories, which is unnecessary.

.gitignore
```diff
@@ -132,3 +132,6 @@ dmypy.json
 
 # include scripts/build directory
 !scripts/build/
+
+# log files
+logs/**
```
twentylemon: Just `logs` should work.

duckbot/__main__.py
Chippers255: There is more than enough space on the SD card so 250kb is fine.

duckbot/__main__.py
```diff
@@ -56,6 +58,20 @@ def intents() -> Intents:
     return intent
 
 
+def logger_setup():
+    log_directory = "logs"
+    if not os.path.exists(log_directory):
+        os.mkdir(log_directory)
```
twentylemon: Minor, [os.mkdirs](https://docs.python.org/3/library/os.html#os.makedirs) has `exist_ok` so you could skip the `if`

jameshughes89: Good good

twentylemon (approved):
Looks good to me, brother.

WELCOME TO THE TEAM 🚀

PR #214, opened 2021-05-16
Retrieve archive of logs with `!logs` command

### LINKED ISSUE #210: `!logs` to attach logs to a discord message
At some point #208 will get merged in and we'll have logs. The duckbot instance is unreachable for most of us, so it'd be nice to have a way for anyone to fetch the logs when stuff goes wrong. And then, hopefully `!logs` never fails.

[Context#send](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Context.send) accepts up to 10 files to send. We could either zip all the logs up, or just send all of the raw files.

jameshughes89:
`!logs` will return an archive of the logs. close #210 

Originally this was setup under cogs, but has been moved out to it's own category --- _logging_.

Eventually the setup for the logs and for overriding the `on_command_error` (to include exceptions in the duck.logs) will be moved to logging.


duckbot/logging/logs.py
```diff
@@ -0,0 +1,23 @@
...
+        self.log_dir = os.path.join(self.duck_dir, "logs")
+        self.log_archive_path = os.path.join(self.duck_dir, "logs.tar.gz")
+        self.compress_command = ["tar", "-zc", "-f", self.log_archive_path, self.log_dir]
+
+    @commands.command(name="logs")
+    async def logs_command(self, context):
+        await self.__logs(context)
+
+    async def __logs(self, context):
+        subprocess.call(self.compress_command)
```
twentylemon: Strongly prefer to stay in python whenever possible, `subprocess` makes assumptions about the state of the machine. For compressing the logs, python has [tarfile](https://docs.python.org/3.8/library/tarfile.html) or [zipfile](https://docs.python.org/3.8/library/zipfile.html) built-in.

duckbot/logging/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .logs import GetLogs
+
+
+def setup(bot):
+    bot.add_cog(GetLogs(bot))
```
twentylemon, line 5: This function is not tested. You can more or less copy one of the existing `test_setup.py` files from another extension.

tests/logging/test_logs.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@pytest.mark.asyncio
+async def test_archive_sent(bot, context):
+    clazz = GetLogs(bot)
+    await clazz._GetLogs__logs(context)
+
+    assert "file" in context.send.call_args.kwargs
+    assert isinstance(context.send.call_args.kwargs["file"], discord.file.File)
+    assert context.send.call_args.kwargs["file"].filename == "duck.logs"
```
twentylemon: Prefer to make as few assertions as possible in your tests, and make the assertion itself contain as much info as possible. When the tests fail, we want to know why without having to understand an assertion failure like _expected true but got false_.

Ideally here, you're only make one assertion, `context.send.assert_called_once_with(file=file)`. Though, it seems `discord.File` doesn't implement `==`, so we'll need a few in this case, but we can still improve the errors we'd get here. Maybe something like:
```python
context.send.assert_called_once()
assert context.send.call_args.kwargs["file"].filename == "duck.logs"
```

But, I'm not a big fan of that either. We could define our own equals, similar to how we did for [Embed](https://github.com/Chippers255/duckbot/blob/main/tests/conftest.py#L144-L160) (see #177). Patching in `==` and `str()` for file would allow use to do the much cleaner `context.send.assert_called_once_with(file=file)`, but comes with the caveat that we have to write `==` and `str()` for a class we don't own. But, `==` is fairly simple, instance check and check that `fp` and `filename` are equal; and `str()` is `str(join(fp, filename))`.

tests/logging/test_logs.py
```diff
@@ -0,0 +1,21 @@
+import discord
+import os.path
+import pytest
+from duckbot.logging import GetLogs
+
+
+@pytest.mark.asyncio
+async def test_archive_sent(bot, context):
```
twentylemon: Test names should be of the format `test_<unit-under-test>_<expectation>`. Eg, this one could be `test_logs_sends_tarball_of_logs`.

duckbot/logging/logs.py
```diff
@@ -0,0 +1,23 @@
...
+        log_archive = discord.File(self.log_archive_path, "duck.logs")
+        await context.send(file=log_archive)
+        subprocess.call(["rm", self.log_archive_path])
```
twentylemon: Is this required (removing the tarball)? The logs are only ~4mb, the tarball is likely to be like ~5% of that size. Will subsequent `!logs` invocations overwrite the tarball?

duckbot/logging/logs.py
```diff
@@ -0,0 +1,23 @@
+import discord
+from discord.ext import commands
+import os
+import subprocess
+
+
+class GetLogs(commands.Cog):
```
twentylemon: Please rename the file to match the sole symbol defined in it. `get_logs.py` or rename this class to `Logs`. If you change the file name, change the test file name to match as well.

tests/logging/test_logs.py
```diff
@@ -0,0 +1,21 @@
...
+    assert "file" in context.send.call_args.kwargs
+    assert isinstance(context.send.call_args.kwargs["file"], discord.file.File)
+    assert context.send.call_args.kwargs["file"].filename == "duck.logs"
+
+
+@pytest.mark.asyncio
+async def test_archive_removed(bot, context):
+    clazz = GetLogs(bot)
+    await clazz._GetLogs__logs(context)
+    assert not os.path.exists(os.path.join(os.getcwd(), "logs.tar.gz"))
```
twentylemon: This `os` interactions need to be mocked out. Tests shouldn't have side effects!

twentylemon (changes requested):

twentylemon:
Did you perform any manual testing? Can you include some examples for posterity?

tests/logging/test_logs.py
twentylemon: > But, I'm not a big fan of that either.

I should be clear, the `eq` and `str` changes are larger, and maybe beyond the scope of this PR. Could be a follow up. For this making better assertions is good enough, though I won't poo-poo if you opt for the larger change.

duckbot/logging/logs.py
jameshughes89: I flip-flopped. No real preference either way. More did it to play with tests.

duckbot/logging/logs.py
jameshughes89: Done. Also, switched to `os.remove`.

duckbot/logging/__init__.py
jameshughes89: Done.

tests/logging/test_logs.py
jameshughes89: Done.

duckbot/logging/logs.py
jameshughes89: Done.

duckbot/logging/logs.py
jameshughes89: I opted to not remove the archive and cut the functionality.

tests/logging/test_logs.py
jameshughes89: Irrelevant now as the remove has been cut.

tests/logging/test_logs.py
jameshughes89: Changed to suggested assertion. 

RE: defining our own `== ` and `str()`, where do we feel is the best place to write this?

jameshughes89: I assumed `fp` was a filepath, but it's actually an `_io.BufferedReader`. 

I do not believe we can easily compare `fp` values. We can compare `a.fp.read() == b.fp.read()` however, but that may be overkill. 

Below is an example:
```
In [4]: a = discord.file.File('logs.tar.gz')

In [5]: b = discord.file.File('logs.tar.gz')

.
.
.

In [10]: a.fp
Out[10]: <_io.BufferedReader name='logs.tar.gz'>

In [11]: type(a.fp)
Out[11]: _io.BufferedReader

In [12]: a.fp == b.fp
Out[12]: False

.
.
.

In [31]: a.fp.read() == b.fp.read()
Out[31]: True

```

twentylemon: That's annoying, I expected it to be a path-like object (like fp -> filepath). Reading the stream won't make sense in tests since the underlying file may not exist. In which case, I guess we can pass on defining `==` for now.

jameshughes89: If we _really_ wanted, I suppose just compare file name attribute for `==` since that's all that we're currently doing in the test.

tests/logging/test_get_logs.py
```diff
@@ -0,0 +1,19 @@
+from unittest import mock
+
+import pytest
+
+from duckbot.logging import GetLogs
+
+
+@pytest.mark.asyncio
+@mock.patch("discord.File", audospec=True)
+@mock.patch("tarfile.open", autospec=True)
```
twentylemon: Can probably remove the `autospec` -- or at least fix `audospec`. `autospec` seems to me to be more for when you _use_ the mock a lot more; care about accessing fields that don't exist or you're stubbing a lot of behaviour. None of that is really happening here.

tests/logging/test_get_logs.py
```diff
@@ -0,0 +1,19 @@
...
+async def test_get_logs_sends_tarball_of_logs(mock_open, mock_dis_file, bot, context):
```
twentylemon: I'd drop `mock` from the names, the context already implies they are mocks. And `dis` means nothing, so just drop that too. `tar` or `open` and `file` are sufficient here.

tests/logging/test_get_logs.py
```diff
@@ -0,0 +1,19 @@
...
+    mock_open.add = mock.MagicMock()
```
twentylemon: This isn't quite right. You're stubbing a function, you don't want to add `tarfile.open.add`; rather, you want to change the `return_value` of `open`. Further, you should verify how this mock was interacted with. Your test now only checks `context.send` was called with a `discord.File`.

More correctly;

```python
@mock.patch("tarfile.open")
@mock.patch("tarfile.Tarfile") # or whatever type open returns
# rest left out for brevity
async def test(tar, open):
    open.return_value = tar
    # call !logs ; assert context.send ; essentially your existing test
    open.assert_called_once_with(log_dir, "w:gz")
    tar.add.assert_called_once_with(the_logs_directory)
    tar.close.assert_called_once()
```
The above verifies that a tarfile was created using the `log_dir`, in addition to sending the file.

_I think_ you could also even verify how the `discord.File` was constructed, which would pretty much capture everything going on in the unit under test.
```python
file.assert_called_once_with(tar_location, name)
```

twentylemon (changes requested):
Mocks! Huzzah!

Chippers255:
This should return a clean archive. It currently returns the entire path to the logs. For example: `logs.tar.gz\home\jhughes\Insync\jameshughes89@gmail.com\Google Drive\codeProjects\duckBot\logs`

tests/logging/test_get_logs.py
jameshughes89: Made it discord_file instead of file because file is a type. Other than that, done.

tests/logging/test_get_logs.py
jameshughes89: i like audospec. But done.

twentylemon (approved):
He has done it.

🚀

PR #219, opened 2021-05-17
Add a random `Game` that Duckbot is playing when it starts

Chippers255:
When DuckBot is started it will now select a random `Game` to be playing. I have added a few initial choices as well as Gloomhaven to get Dave's rocks off. Duck Game is still an option.


twentylemon (changes requested):
I like the idea! Please update the tests -- they currently check that the activity is _Duck Game_.

twentylemon (approved):

PR #223, opened 2021-05-18
lazily construct db client to remove need for libpq in tests

### LINKED ISSUE #211: remove need for `ffmpeg` and postgres libraries for running tests
Currently, audio tests fail if `ffmpeg` isn't installed (it is used by discord-py internally), and anything using the database fails if the postgres drivers are not installed (`libpq-dev`). These dependencies for test-time are just annoying, we should remove them.

twentylemon:
Fixes #211 when paired with #222. ~~This is a draft for now, waiting on #222 to be merged in.~~

This one is similar to #222, in that [where to patch](https://docs.python.org/3/library/unittest.mock.html#where-to-patch) was a part of it. I also changed it so the db client is lazily constructed, other tests were creating real instances of the `Database`, so it just means I don't have to change those tests.


jameshughes89 (commented):
```
@mock.patch("duckbot.db.database.create_engine")
@mock.patch("tests.db.test_database.Base")
def test_session_creates_tables(create, base):
```

I think the order may need to be reversed. [See the first note](https://docs.python.org/3/library/unittest.mock.html#quick-guide)

twentylemon:
> ```
> @mock.patch("duckbot.db.database.create_engine")
> @mock.patch("tests.db.test_database.Base")
> def test_session_creates_tables(create, base):
> ```
> 
> I think the order may need to be reversed. [See the first note](https://docs.python.org/3/library/unittest.mock.html#quick-guide)

Lol. I never saw that note, I've actually been writing tests assuming the opposite for months. Should probably do a once over to fix names through the tests. I'll fix these ones.

It doesn't actually affect the behaviour of tests though. In this case, what's important is the two classes are mocked out, the names assigned to the patches in the function doesn't matter since they aren't used otherwise.  
In most other places where [mock.patch is used](https://github.com/Chippers255/duckbot/search?q=%40mock.patch), the same principal applies -- they don't actually affect tests, but we should probably change them.

Chippers255 (approved):

PR #224, opened 2021-05-18
correct test argument names when using mock.patch

twentylemon:
As @jameshughes89  pointed out in #223, mock.patch decorators apply in reverse-ish order. This just corrects test argument names through all the tests. No behaviour was broken before, since we either didn't use the patched class directly, or added the behaviour we wanted to the mock anyways. This is essentially just for correctness, to prevent confusion in the future.

For finding this, I just did `grep -rn 'mock.patch' *` and looked for two mock.patches on consecutive lines. There's surprisingly few, probably because we use fixtures for most mocks.


jameshughes89 (approved):
Looks good. Easy switcharoo.

PR #226, opened 2021-05-20
Move logging setup into its own file.

jameshughes89:
Moved the logging setup function to its own file within logging. Added tests for logging setup.


duckbot/__main__.py
```diff
@@ -49,19 +48,7 @@ def run_duckbot(bot: commands.Bot):
...
-    logger.setLevel(logging.INFO)
-
-    handler = logging.handlers.RotatingFileHandler(filename=os.path.join(log_directory, "duck.log"), mode="a", maxBytes=256000, backupCount=10)
-    handler.setFormatter(logging.Formatter("%(asctime)s:%(levelname)s:%(name)s: %(message)s"))
-    logger.addHandler(handler)
-
-
 if __name__ == "__main__":
-    logger_setup()
+    duckbot.logging.define_logs.define_logging()
```
twentylemon: Did you consider doing the logging setup as part of the logging extension `setup` method?

duckbot/cogs/announce_day/phrases.py
```diff
@@ -1,74 +1,11 @@
...
-            "WEEKEND ALERT: {0}!",
-        ],
-    },
+    0: {"names": ["the day of the moon", "Moonday", "Monday"], "templates": ["Screw this, it is {0}."]},
+    1: {"names": ["Tyr's day", "Tuesday", "Dndndndndndndnd", "Taco Tuesday"], "templates": ["Yo, one of my favourite days ever: {0}."]},
+    2: {"names": ["Odin's day", "Wednesday", "Wednesday, my dudes", "hump day", "Ness' wedding day", "https://www.youtube.com/watch?v=du-TY1GUFGk"], "templates": []},
+    3: {"names": ["Thor's day", "Thorsday", "civ day", "Thursday"], "templates": []},
+    4: {"names": ["Frigga's day", "Friday, Friday, gotta get down on Friday", "Stir Fry day", "Friday"], "templates": []},
+    5: {"names": ["Tom's Movie Night", "Saturn's day", "Saturday"], "templates": ["WEEKEND ALERT: {0}!"]},
+    6: {"names": ["the day of the sun", "Sunday", "Sunday, Sunday, Sunday"], "templates": ["WEEKEND ALERT: {0}!"]},
```
twentylemon: Revert these changes. Occurs in several files. Something on your system seems misconfigured, `black` won't remove trailing commas in lists and such.

tests/logging/test_define_logs.py
```diff
@@ -0,0 +1,26 @@
...
+@mock.patch("os.makedirs")
+async def test_define_logs_create_logger(make_dirs, get_logger, logger, handler):
+
+    get_logger.return_value = logger
+
+    define_logging()
+
+    make_dirs.assert_called_once_with("logs", exist_ok=True)
+    get_logger.assert_called_once_with("discord")
+    logger.setLevel.assert_called_once_with(20)  # 20 corresponds to INFO
```
twentylemon: Still use `logging.INFO` instead of `20`. `20` is a python internal implementation detail we don't really have to care about. What's important here is `INFO` is used.

tests/logging/test_define_logs.py
```diff
@@ -0,0 +1,26 @@
...
+    handler.called_once_with("logs", exist_ok=True)
+
+    handler.called_once_with(filename=os.path.join("logs", "duck.log"), mode="a", maxBytes=256000, backupCount=10)
+    logger.addHandler.called_once_with(handler)
```
twentylemon: Should all be `assert_called_once_with` ?

twentylemon (changes requested):

duckbot/__main__.py
jameshughes89: I did, however I was unclear on how/when the setup was actually called.

tests/logging/test_define_logs.py
jameshughes89: absolutely they should

duckbot/__main__.py
twentylemon: It's called by `load_extension`. A discord-py thing.

jameshughes89: Based on that, I feel compelled to include it in setup. Let me know if y'all agree.

duckbot/cogs/announce_day/phrases.py
jameshughes89: Done.

tests/logging/test_define_logs.py
jameshughes89: Done.

duckbot/__main__.py
jameshughes89: Since I will be ripping [this off to overwrite the `on_command_error`](https://gist.github.com/EvieePy/7822af90858ef65012ea500bcecf1612), I will have that dealt with within the logging setup. 

With this, the setup within logging will:
(a) Setup the logging 
(b) deal with the `on_command_error`
(c) deal with get logs

Since the setup is _setting up_ the other bits, I can see an argument to add it here. Although I think we can just as easily move the `duckbot.logging.define_logs.define_logging()` to be called in the setup and leave the function it in it's own file, all the other things are classes and the `define_logging()` is just a function. 

Leave open until we come to a good enough reason either way.

twentylemon: The `duckbot.logging` module contains stuff related to **all** logging within duckbot -- which includes the discord bot specific things like `!logs` and setting the log level for logs from the `discord` module. For `!logs` and eventual event listeners specifically, it makes a lot of sense to have those be setup by the extension `setup` method, since they are discord bot specific (same as all other extension `setup` methods).

I guess for the other bit, the question becomes, are logging behaviours a _bot extension_, or are they something that should be inherent to duckbot? Probably the latter. In which case, I think this is probably the better approach. Plus, as we further define more logs (say, `duckbot` specific ones), it makes sense to place that setup within the `define_logging` function.

However! Please don't refer to it as `duckbot.logging.define_logging.define_logging`. Elevate the function to be a member of the module. In `__init__.py` in the logging module, add:
```python
from .define_logging import define_logging
```
`__init__.py` is executed whenever you import a module. The net affect here is creating `duckbot.logging.define_logging` instead. Which means, you're more free to change your internals (like rename the file) without affecting other code already using it.

twentylemon (approved):
Pog.

PR #231, opened 2021-05-21
Record Exceptions in Logs

### LINKED ISSUE #212: ensure execptions are logged
Logs were added in #208, but apparently discord-py only dumps exceptions to stderr by default. See [Bot#on_command_error](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#discord.ext.commands.Bot.on_command_error) and [Client#on_error](https://discordpy.readthedocs.io/en/stable/api.html#discord.Client.on_error). The docs mention to override it, which we can do to cram traces into the logs. I think also dumping to stderr still makes sense, but we definitely want exceptions in logs.

jameshughes89:
Record exceptions from commands in the log files. Currently this does **not** include any `on_message` exceptions, which I assume is possible, but that is outside the scope of this PR. Has to do with #212


duckbot/logging/exception_logs.py
```diff
@@ -0,0 +1,27 @@
...
+from discord.ext import commands
+
+
+class ExceptionLogs(commands.Cog):
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.Cog.listener()
+    async def on_command_error(self, context, exception):
+        await self.__on_command_error(context, exception)
```
twentylemon: The private method thing here isn't required, you can invoke `@listener` methods directly in your tests. Also, imo prefer changing the name to describe the method's purpose, rather than the event it responds to. Ie:
```python
@commands.Cog.listener("on_command_error")
async def log_exceptions(self, context, exception):
    # blah
```

duckbot/logging/exception_logs.py
```diff
@@ -0,0 +1,27 @@
...
+
+    async def __on_command_error(self, context, exception):
+        if hasattr(context.command, "on_error"):
+            return
+        cog = context.cog
+        if cog and cog._get_overridden_method(cog.cog_command_error) is not None:
+            return
```
twentylemon: You're calling private methods here, definitely should not be doing that. Can you delegate this behaviour to `self.bot.on_command_error` maybe?

Do we ever really want to bail early from this? All we're doing is logging the exception. Seems to be we'd _always_ want that to happen.

twentylemon (changes requested):
What's the behaviour for `Cog.on_command_error`? Does it only respond to command errors in that cog? Or from the entire bot? I'm wondering if this new cog could simply add a listener to bot instance (similar to [on_ready](https://github.com/Chippers255/duckbot/blob/main/duckbot/duckbot.py#L24-L27)) which then applies to all command errors.

I'd also say this _does not_ fix #212. It's one part of it, we definitely want exceptions from events like `on_message` to be logged, as well as exceptions from `@task` methods.

jameshughes89:
Based on tests, it responds to all command errors. 

Will look into the bot idea.

duckbot/logging/exception_logs.py
jameshughes89: This is the overwritten method copy/pasted with two lines added for the logger. 

Would we ever want a specific error handler for a specific cog or something that has different behavior? Would we ever not want to log the exception?

twentylemon: I can't see any reason why we'd ever want to not log an exception. Invoking a cog specific error handler can still take place (presumably, that can be handled by the original method), but we may as well log all exceptions here so we don't have to remember to anywhere else.

jameshughes89:
Although having it as a cog listener had it work for all cogs, I made it a listener for the whole bot. I was hoping this would magically make it work with messages too, but that was not the case.

twentylemon:
https://discordpy.readthedocs.io/en/latest/api.html#discord.on_error
> on_error will only be dispatched to Client.event().
> 
> It will not be received by Client.wait_for(), or, if used, Bots listeners such as listen() or listener().

So `listener` won't work (that's... odd). But seems `event` should? Could maybe try adding in `setup`:

```python
@bot.event
async def on_error(*args):
    log_it_bro()
```

twentylemon:
> Although having it as a cog listener had it work for all cogs, I made it a listener for the whole bot. I was hoping this would magically make it work with messages too, but that was not the case.

If it didn't work, I'd probably flip back to `@listener` then, just to keep the style consistent throughout all the cogs. The only weirdo that needs to use `add_listener` is the bot class itself.

jameshughes89:
> https://discordpy.readthedocs.io/en/latest/api.html#discord.on_error
> 
> > on_error will only be dispatched to Client.event().
> > It will not be received by Client.wait_for(), or, if used, Bots listeners such as listen() or listener().
> 
> So `listener` won't work (that's... odd). But seems `event` should? Could maybe try adding in `setup`:
> 
> ```python
> @bot.event
> async def on_error(*args):
>     log_it_bro()
> ```

Tried to hack it in there with no success. 

I recommend addressing the `on_message` and `task` in a seperate PR as it seems to need some non insignificant tinkering.

duckbot/logging/exception_logs.py
```diff
@@ -0,0 +1,18 @@
...
+class ExceptionLogs(commands.Cog):
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.Cog.listener("on_command_error")
+    async def log_exceptions(self, context, exception):
+        logger = logging.getLogger("discord")
+        exception_string = "".join(traceback.format_exception(etype=type(exception), value=exception, tb=exception.__traceback__))
+        logger.error(exception_string)
+        print("Brother, ignoring exception in command {}:".format(context.command), file=sys.stderr)
```
twentylemon: Prefer f-strings to explicit `format` calls. This can be `f"Brother, ignoring exception in command {context.command}:"`

tests/logging/test_exception_logs.py
```diff
@@ -0,0 +1,28 @@
...
+@mock.patch("traceback.format_exception")
+@mock.patch("logging.Logger")
+@mock.patch("logging.getLogger")
+async def test_exception_logs_log_exception(get_logger, logger, format_exc, print_exc, bot, context):
+    exception = mock.Mock(Exception)
+    cog = mock.Mock(commands.Cog)
+    get_logger.return_value = logger
+    context.command = ""
+    context.cog = cog
+    context.cog._get_overridden_method.return_value = None
```
twentylemon: Remove cog references? These seem like leftovers.

twentylemon (approved):

tests/logging/test_exception_logs.py
jameshughes89: Good catch. These were used for ensuring no early bailing on functio being tested.

duckbot/logging/exception_logs.py
jameshughes89: Agreed. Changed.

PR #233, opened 2021-05-22
add duckbot module logs to log files

twentylemon:
Also adding logs from the `duckbot` module into the logs files. We don't log anything yet, but we may in the future, so just setting that up.


jameshughes89 (approved):
Good general cleanup of the code too. <3

PR #234, opened 2021-05-23
add jetbrains idea to gitignore

twentylemon:
I'm fiddling locally with pycharm just to see how it fairs. I've added the global jetbrains gitignore from [gitignore](https://github.com/github/gitignore), and added the whole `.idea` for now while I see if I like it or not. I may just leave it there forever, who knows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+# log files
+logs
+logs.tar.gz
+
+##### https://github.com/github/gitignore/blob/master/Python.gitignore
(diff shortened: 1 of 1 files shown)


jameshughes89 (approved):
Looks good. Only Q is about the last line --- is that only for Android studio?

twentylemon:
> Looks good. Only Q is about the last line --- is that only for Android studio?

Don't know don't care. It's from the gitignore repo.

PR #239, opened 2021-05-25
rename logging module to logs to avoid name clash with builtin; duckbot file with main module

twentylemon:
Intellij was freaking out over the `logging` module, it keeps getting it confused with the builtin python one. Simply renaming to `logs` to avoid the name clash.

Similarly, renaming the `duckbot.py` file to just `bot.py`.


jameshughes89:
One potential issue is, the logs directory for the actual logs is `./logs/`, and now we will have `./duckbot/logs/`. The `./logs/` is not included in git, but will exist once an instance of duckbot runs.

twentylemon:
> I potential issue is, the logs directory for the actual logs is `./logs/`, and now we will have `./duckbot/logs/`. The `./logs/` is not included in git, but will exist once an instance of duckbot runs.

gitignore paths are relative when you specify `./thing`, so `./logs` refers only to `<root>/logs`, not `<root>/duckbot/logs`. I didn't force add the `duckbot/logs` directory here, just `git add .` captured them.

jameshughes89:
I didn't mean so much about `.gitignore` as using the same directory name so close together.

jameshughes89 (approved):
I am pretty sure that this is all references to `logs` & `logging`.

PR #246, opened 2021-05-26
hacky support for discord slash commands

### LINKED ISSUE #200: use slash commands instead of `!` commands
Discord recently released access to slash commands for bots. They're a lot more intuitive -- you can hit `/` and see the list of commands available, including the arguments and a short description.

Unfortunately, discord py won't be supporting them until a 2.0 release, see https://github.com/Rapptz/discord.py/issues/6149

I did find [discord-py-slash-commands](https://discord-py-slash-command.readthedocs.io/en/latest/) though, which may be a decent replacement in the meantime.

twentylemon:
Fixes #200 

Oh boy. Where do I even start with this one. With this, duckbot has some hacky support for discord's built-in slash commands. I do suggest reading the [discord docs](https://discord.com/developers/docs/interactions/slash-commands) on this one as you read through the code. It'll help put things into context.

For this, I wanted to more or less write an adapter between slash commands and discordpy commands. We already have many commands that I don't want to have to rewrite, so I instead intercept the slash commands and delegate to the existing commands. The overall idea is to create an `InteractionContext` class which behaves _nearly_ the same as the discordpy `Context` class, except it interacts with slash command stuff instead. In this way, we'd be able to use literally the same code for both command variants.

So, in some details.

`@slash_command` is a decorator for existing `@command` methods, which will also register them as a slash command. It basically crams new fields into the discordpy commands, which I use later when slash commands are invoked.

`SlashCommandPatch` is the cog which creates/updates/deletes slash commands that were registered via `@slash_command`. It also hacks into discordpy to create an `on_slash` event, which the cog registers a listener for. That `on_slash` effectively creates some beans for the data sent by discord, then throws those at the associated discordpy command.

For tests, I changed the `context` fixture to be _both_ discordpy contexts and the custom interaction contexts. It should help us capture errors (like stuff we haven't implemented) early. I still kept around `command_context` for discordpy-only commands.


twentylemon:
I may discard this. I'll have to revisit. With stuff like threads being added to discord, I'm not sure how much _we_ want to be the ones supporting changes.

PR #250, opened 2021-05-28
add default codecov config

twentylemon:
codecov seems to be broken since [config was added](https://github.com/Chippers255/duckbot/commit/74c26faa3d0aafcac1c66f7ee1f35e72e8b73c8c), not sure why. This is just re-adding the [default config](https://docs.codecov.io/docs/codecov-yaml), in addition to the threshold configs added into mentioned commit.


twentylemon:
DIDN'T WORK

PR #257, opened 2021-05-30
add `!ascii` command to convert text into ascii art

twentylemon:
I mean. Title. See beta #dank channel for some examples.


jameshughes89 (commented):
Should there be a test for no text?

twentylemon:
> Should there be a test for no text?

Eh. There's a default argument for the command method, and the delegate accepts that. I could test receiving none in the delegate, or just leave it as undefined behaviour since we don't expect it. I'm fine with leaving it undefined, but I'm not wholly against adding the test if you feel strongly.

jameshughes89 (approved):

PR #264, opened 2021-06-01
yeet some messages to keep chats a little cleaner

### LINKED ISSUE #147: yeet command messages to keep chats a little cleaner
[Message#delete](https://discordpy.readthedocs.io/en/latest/api.html#discord.Message.delete) is a thing. We could globally catch commands and delete the source message to keep the clutter down.

twentylemon:
Fixes #147. I opted not to delete most command messages since they kinda serve as documentation. We'd have to change the responses to also include the question and who asked it (like a `!weather` response by itself seems weird).

I did opt to delete some error messages after a short time though.

### DIFF
--- a/duckbot/cogs/duck/duck.py
+++ b/duckbot/cogs/duck/duck.py
@@ -3,6 +3,7 @@ import random
 from discord.ext import commands
 
 from duckbot.util.emojis import regional_indicator
+from duckbot.util.messages import try_delete
 
 
 class Duck(commands.Cog):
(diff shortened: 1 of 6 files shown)


duckbot/cogs/duck/duck.py
```diff
@@ -23,15 +24,20 @@ async def react_with_duckbot(self, message):
                 await message.add_reaction(regional_indicator(c))
 
     @commands.command(name="duck")
-    async def github(self, context):
-        await self.__github(context)
+    async def duck_command(self, context):
+        await self.github(context)
 
-    async def __github(self, context):
+    async def github(self, context):
```
jameshughes89, line 30: Why change from `__github`? Why not `_github`?

jameshughes89 (approved):
Sufficient yeeting

duckbot/cogs/duck/duck.py
twentylemon: Eh. More or less, I just find it cleaner, easier to read this code and the tests. As for private-ness, `_github` is no better than `github`, I don't really see the point.

If we get to a point where commands are actually interconnected (one command invokes another, or something), then we can revisit to "hide" these implementations, but for now. I like readability.

PR #265, opened 2021-06-01
add towel day to list of special days

Chippers255:
We completely forgot about towel day this year, this change should stop that happening in future.

I literally did not check that this change didn't break anything. But based on the other examples, everything should be fine. After all, DON'T PANIC.


twentylemon (approved):
:towel:

PR #268, opened 2021-06-03
Revert "install `libpulse-dev` for matplotlib (#266)"

twentylemon:
This reverts commit 7c58724f

Doesn't seem like it's needed. Needed for PIs, but docker != pi.


jameshughes89 (approved):
Looks reverted to me.

PR #273, opened 2021-06-04
use timezonefinder lib instead of tzwhere

twentylemon:
`tzwhere` is taking a lot of time during `!weather`. In my local tests, `!weather` would take like 5s. After this change, it took about 2s.

See [their docs](https://timezonefinder.readthedocs.io/en/latest/3_about.html#comparison-to-pytzwhere) for a comparison between the two libs.


jameshughes89 (approved):
2s is better than 5s.

PR #277, opened 2021-06-05
remove `--no-cache-dir` option to avoid double building some packages

twentylemon:
I compared a [recent main build](https://github.com/Chippers255/duckbot/runs/2752505186) to actions from my PR.

* [action](https://github.com/twentylemon/duckbot/runs/2752968256) - expand "Build and Push Docker Image" then search for "cached"
    * [`numpy` was cached](https://github.com/twentylemon/duckbot/runs/2752968256?check_suite_focus=true#step:7:1675), saving [quite a long build](https://github.com/Chippers255/duckbot/runs/2752505186?check_suite_focus=true#step:7:1721)
        * not entirely sure why that happened, `numpy` may be getting built during the [matplotlib download](https://github.com/Chippers255/duckbot/runs/2752505186?check_suite_focus=true#step:7:1560), which would explain why it takes ~30min to download ~30mb
    * many packages still do have to be built though

This will make the docker layers cache a little bigger, but should save maybe 10-15min per deploy action.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -9,54 +9,54 @@ on:
     - main
 
 jobs:
-  release:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
-      with:
-        python-version: '3.8'
+#   release:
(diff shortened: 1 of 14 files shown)


jameshughes89 (approved):
I found other dudes on the internet suggesting this in the context of docker images.

Issue #247, opened 2021-05-28
add italics/bold to text generated by `!mock`


twentylemon:
MoCkInG tExT is hard to type, a command to automate it would be swell. DuckBot should probably deleting the originating command message for added oomph.

Alternating the case is a good place to start, I also think adding varying **bold** and _italics_ would also be a nice touch.

jameshughes89:
Which cog you thinking to put this under? `text`, with `ascii_art`?

twentylemon:
I'd probably do a new cog in the `text` extension. It's not really related to ascii art.

jameshughes89:
As of now, the TODO for this is _italics_ and **bold**.

> Alternating the case is a good place to start, I also think adding varying bold and italics would also be a nice touch.

jameshughes89:
I almost cut the _new command_ label too, but opted to keep it given that it is still about the new command. 

I could totally see it being removed however.

PR #279, opened 2021-06-05
`!mock` command to write mOcKiNg TeXt

jameshughes89:
Can now use `!mock` to turn text into MoCkiFiEd StUfF. 
`!mock` with no message mocks the called
`!mock` with too much text lets us know it's got too much stuff

This is about #247, but as of now there is no _italics_ or **bold** implemented because testing seems too hard.

### DIFF
--- a/duckbot/cogs/text/mock_text.py
+++ b/duckbot/cogs/text/mock_text.py
@@ -13,14 +13,12 @@ class MockText(commands.Cog):
 
     async def mock_text(self, context, text: str):
         if text == "":
-            mocked_text = await self.mockify("based on this, I should mock you... I need text dude.")
-            await context.send(f"{context.message.author.display_name}, {mocked_text}")
+            mocked_text = await self.mockify(f"{context.message.author.display_name}, based on this, I should mock you... I need text dude.")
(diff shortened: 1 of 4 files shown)


duckbot/cogs/text/mock_text.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @commands.command(name="mock")
+    async def mock_text_command(self, context, *, text: str = ""):
+        await self.mock_text(context, text)
+
+    async def mock_text(self, context, text: str):
+        if text == "":
+            mocked_text = await self.mockify(f"{context.message.author.display_name}, based on this, I should mock you... I need text dude.")
+        elif len(text) > 1990:
+            mocked_text = await self.mockify("that's too much letters and stuff guy.")
```
twentylemon: I don't think we need this check. This restriction is for the 2000 character limit for a discord message, but the command is already a discord message and the mOcKeD vErSiOn is the same length as the input.

twentylemon (approved):

jameshughes89:
Good call on rename btw.

twentylemon (approved):

PR #283, opened 2021-06-07
move wiki contents into this repo; publish wiki on push

twentylemon:
Moving the wiki pages to be part of this repo so they have a chance of staying up to date with the repo itself!

The wikis are taken directly from the existing wiki. No changes there.

The wiki publish uses [a third party action](https://github.com/SwiftDocOrg/github-wiki-publish-action). It requires a github access token with `repo` access, we already have one for triggering the deployment in the deployment repo, the permissions required are the same. I ended up re-using that one.  
I debated updating the wiki repo manually (ie `git@github.com:Chippers255/duckbot.wiki.git`), but it's a little annoying with potential merge conflicts and stuff. Action was pain free, so eh, going with that.

`WIKI_COMMIT_MESSAGE` is [undocumented](https://github.com/SwiftDocOrg/github-wiki-publish-action/blob/v1/entrypoint.sh#L38) but would be the commit message. If not provided, it gives the useless message "Automatically publish wiki." I jam in the same commit message from the push to duckbot instead. See [push event docs](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push).

See a [pared down action](https://github.com/twentylemon/duckbot/runs/2768961827?check_suite_focus=true) I ran against my repo and the [wiki results](https://github.com/twentylemon/duckbot/wiki) from that. Commit messages didn't work there because the pull request event doesn't include the commits.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -9,57 +9,81 @@ on:
     - main
 
 jobs:
-#  release:
-  #    runs-on: ubuntu-latest
-  #    steps:
-  #    - uses: actions/checkout@v2
-  #    - name: Set up Python 3.8
-  #      uses: actions/setup-python@v2
-  #      with:
-  #        python-version: '3.8'
-  #
-  #    - name: Setup Virtual Environment Cache
-  #      id: venv
-  #      uses: actions/cache@v2
-  #      with:
-  #        path: venv
-  #        key: ${{ env.pythonLocation }}-venv-${{ hashFiles('setup.py') }}
-  #
-  #    - name: Install Packages
-  #      if: steps.venv.outputs.cache-hit != 'true'
-  #      run: |
-  #        python -m venv --clear venv
-  #        . ./venv/bin/activate
-  #        pip install --upgrade pip setuptools wheel
-  #        pip install --editable .[dev]
-  #    - name: Run Tests
-  #      run: |
-  #        . venv/bin/activate
-  #        pytest
-  #    - name: Report Code Coverage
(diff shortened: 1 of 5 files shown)


.github/workflows/python-package.yml
```diff
@@ -85,6 +85,14 @@ jobs:
           --set '*.cache-from=type=local,src=/tmp/.buildx-cache' \
           --set '*.cache-to=type=local,dest=/tmp/.buildx-cache'
 
+    - name: Publish Wiki Documentation
+      uses: SwiftDocOrg/github-wiki-publish-action@v1
+      with:
+        path: wiki
+      env:
+        GH_PERSONAL_ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
+        WIKI_COMMIT_MESSAGE: ${{ github.event.commits[0].message }}
```
jameshughes89, line 94: Only line I could not verify, but I assume you tested this bad boy?

jameshughes89 (approved):

.github/workflows/python-package.yml
twentylemon: > Only line I could not verify, but I assume you tested this bad boy?

Nope. From the description:
> See a pared down action I ran against my repo and the wiki results from that. Commit messages didn't work there because the pull request event doesn't include the commits.

Github seems to just populate the non-existent value with an empty string instead of failing, so the PR gave the horrible "auto publish wiki" commit message. It _should_ work, but I need a real push event, which I didn't test.

PR #292, opened 2021-06-25
correct chance to send a gif of the day to 10%

twentylemon:
##### Summary 

I realized today that the current chance to send a gif is actually 9% instead of 10%, since it relies on the dog not being sent. 90% * 10% is only 9%. So, correcting to the indented 10%.

##### Checklist


jameshughes89 (approved):
I'm at least 11.11% sure that's how the math works, but my brain has been off for a while.

twentylemon:
> I'm at least 11.11% sure that's how the math works, but my brain has been off for a while.

It'd be `0.9 * x = 0.1` is what we're after. 0.9 because that's the change we don't send a dog photo, which needs to occur to send a gif.

PR #310, opened 2021-09-09
upgrade discord.py to 2.0 + support threads

### LINKED ISSUE #302: update discord.py to 2.0 for slash commands and threads support
Discord recently released threads, but our version of discord.py does not support them, commands and such get ignored. The latest version on pypi is 1.7.3, which does not have thread support (see [discord.py release tags](https://github.com/Rapptz/discord.py/tags)). Their main branch does have some thread support already (see https://github.com/Rapptz/discord.py/issues/7060), so just waiting on them to publish a new version to pypi.

No code change should be needed explicitly, so long as the pip/docker caches have been invalidated in the github actions (which they are after one week of inactivity). Simply triggering the deploy action should download the new version -- though we'll also likely want to test it out, and maybe add a new [test context](https://github.com/Chippers255/duckbot/blob/main/tests/conftest.py#L93) instance (or channels, message, etc -- however they implement it).

twentylemon:
##### Summary 

Fixes #302 

There's only one backward incompatible change we had to fix, which is that all `datetime`s returned by messages and such are now timezone aware, which means existing comparisons to non timezone aware objects failed. This only affected the `insights` cog. The only other use of datetime was the day announcement, but it's not related to the discord library.

A minor change, duckbot will react in any `#dank` channel, be it a thread or the existing top level channel. I could do top level channel only, but eh.

I added a `Thread` as a default channel type for all the tests, so using `channel` or `context` will give you a thread in addition to the existing channel types. Also did some renaming, since threads are text channels, `text_context` and `text_message` are changed to `guild_*` instead for clarity.

Ran manual tests in beta. Everything looked good -- I didn't test the insights explicitly, but the first time I boot it up, the task failed due to the timezone issue mentioned above. After fixing, I saw no error messages at least.

Finally, [discord.py](https://github.com/Rapptz/discord.py) is set to read only mode, so building from their source is safe enough, it'll never change. They never published 2.0 to pypi, so we do have to install it from github. 

##### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -52,10 +52,10 @@ async def bot_spy() -> DuckBot:
 
 @pytest.fixture
 @mock.patch("duckbot.DuckBot", autospec=True)
-def bot(b) -> DuckBot:
+def bot(b, monkeypatch) -> DuckBot:
     b.loop = mock.Mock()
-    with mock.patch("discord.ext.tasks.Loop"):  # mock out loop, it uses `asyncio.get_event_loop()` by default
-        return b
+    monkeypatch.setattr(discord.ext.tasks, "Loop", mock.Mock())  # mock out loop, it uses `asyncio.get_event_loop()` by default
+    return b
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -79,11 +79,11 @@ def message(m, channel, user, member) -> discord.Message:
     return m
 
 
-@pytest.fixture
+@pytest.fixture(params=["discord.TextChannel", "discord.Thread"])
 @mock.patch("discord.Message", autospec=True)
-def text_message(m, text_channel, member) -> discord.Message:
-    """Returns a guild TextChannel message with the channel property set."""
-    m.channel = text_channel
+def guild_message(m, request, text_channel, thread, member) -> discord.Message:
+    """Returns a guild TextChannel or Thread message with the channel property set."""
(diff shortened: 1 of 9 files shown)


twentylemon:
Hmm. Lots of warnings in the tests, didn't notice them locally. Will move to draft.

twentylemon:
> Hmm. Lots of warnings in the tests, didn't notice them locally. Will move to draft.

Culprit seems to be bad mocking. Discord.py's loop is hard to mock properly, so `monkeypatch` it is. Replacing the class with a mock removes the warnings. Hackily, but eh, the loops shouldn't be running in tests anyways.

tests/conftest.py
```diff
@@ -163,6 +167,13 @@ def group_channel(g) -> discord.GroupChannel:
     return g
 
 
+@pytest.fixture
+@mock.patch("discord.Thread", autospec=True)
+def thread(thrd) -> discord.Thread:
+    thrd.type = discord.ChannelType.public_thread
+    return thrd
```
jameshughes89, line 174: :hankey:

jameshughes89 (approved):
LGTM (I admit there was some going on faith on the tests, but they passed, so that's cool)

twentylemon:
> LGTM (I admit there was some going on faith on the tests, but they passed, so that's cool)

Spoke offline, this was about the `monkeypatch` for the discordpy `Loop` class. Since the loop is created by decorating a function, it's hard to mock out. The monkeypatch forces that loop to not actually run during the tests, but we can still verify behaviours and such.

PR #314, opened 2021-09-17
speed up unit tests

twentylemon:
##### Summary 

I ran some profiling recently and found that setting up fixtures were taking a huge amount of time in our tests. The main culprit therein was `mock.patch(autospec=True)`. Autospec-ing does runtime analysis of the object being mocked and will throw an error if you try to access fields/methods on the mock that don't exist on the actual object. Pretty useful for tests, but it's very expensive.

With some fiddling, I found you can actually just copy a mock. Copying is still slow, but it's a lot faster than creating a new mock for every test. Here, I cache an instance of every autospec'd mock once, then copy it for tests as they request it.

After that, I profiled more and found some fixtures were being requested (and thus, copied) more often than they should be. The main culprit here was the `channel` fixture, which gives you one of each type of text based channel. The way it was implemented, it'd actually eagerly create _every_ text channel for _every_ text channel -- so, we'd actually create 4x4 channels, but return 4. Thankfully, the pytest `request` fixture has a way to lazily fetch a fixture by name, so I just take advantage of that. This also happens in the `message` fixture, which creates `user` and `member` but only needs one of them per return.

Results are pretty good. On my machine, I got ~35s runs of the whole test suite after all the changes. Before, it was ~80s. My machine is quite a bit faster than the github runners, but we should still expect to see a decent speedup. I tried with and without both sets of changes, they both contribute fairly evenly to the overall speedup.
* [recent main actions run](https://github.com/Chippers255/duckbot/runs/3626983613?check_suite_focus=true) took 2m54s
* [this actions run](https://github.com/Chippers255/duckbot/runs/3628000458?check_suite_focus=true) took 1m42s

I also verified the autospec-ing still works. I added `cow = bot.cow` to a test, and still got: `AttributeError: Mock object has no attribute 'cow'`

##### Checklist


jameshughes89 (commented):
Took too long to connect the dots of why you called the autospec method `.of`

jameshughes89 (approved):
Look at that 1:42 test in release.

PR #318, opened 2021-09-26
automatically disconnect music stream after ~4h

twentylemon:
##### Summary 

Title! Causes `!start` to automatically disconnect after playing the song 75 times, which is about 4h20min. This is just to prevent it from running indefinitely when someone forgets to `!stop`.

I also renamed the methods slightly, so `start` -> `start_command` and the private `__start` -> `start`, similar to other cogs.

Tested manually in beta using `bruh.mp3` and reducing the max count to only 3.

##### Checklist

### DIFF
--- a/tests/cogs/audio/who_can_it_be_now_test.py
+++ b/tests/cogs/audio/who_can_it_be_now_test.py
@@ -75,8 +75,8 @@ async def test_task_loop_repeats_max_times(ffmpeg, vol, bot_spy, context, voice_
     for i in range(76):
         await clazz.stream.wait()
         await asyncio.sleep(0)
+        assert clazz.streaming is (i < 75)  # should be streaming 75 times only
     # stop() is called after song is played 75 times
-    assert clazz.streaming is False
     assert clazz.audio_task is None
     assert clazz.voice_client is None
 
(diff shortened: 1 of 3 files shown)


tests/cogs/audio/who_can_it_be_now_test.py
```diff
@@ -40,20 +40,42 @@ async def test_task_loop_repeats(ffmpeg, vol, bot_spy, context, voice_client, sk
...
+    clazz = WhoCanItBeNow(bot_spy)
+    await clazz.connect_to_voice(context)
+    assert clazz.voice_client is not None
+    await clazz.start(context)
+    assert clazz.audio_task is not None
+    assert clazz.streaming is True
+    for i in range(76):
+        await clazz.stream.wait()
+        await asyncio.sleep(0)
+    # stop() is called after song is played 75 times
```
jameshughes89, line 79: Should there be a streaming check `False` for that 76th run?

twentylemon: Sure, makes sense to assert that it's the 75th one that causes the stop.

jameshughes89 (approved):

twentylemon:
> > The diff coverage is `80.00%`.

Diff coverage is low due to the command method name changes. Commands aren't directly covered by tests.

PR #323, opened 2021-09-30
change docker image to `duckdynasty/duckbot`

twentylemon:
##### Summary 

DuckBot has an org now, I've gone ahead and made an org specific docker account for it. These two instances of `chippers` seem to be the only ones, see [search](https://github.com/duck-dynasty/duckbot/search?q=chippers255).

https://hub.docker.com/repository/docker/duckdynasty/duckbot -- note that docker for some reason doesn't allow dashes in their user names, so I couldn't do `duck-dynasty`.

I'll update the `DOCKER_PASS` secret after approval.

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -75,7 +75,7 @@ jobs:
     - name: Login to Docker
       uses: docker/login-action@v1
       with:
-        username: chippers255
+        username: duckdynasty
         password: ${{ secrets.DOCKER_PASS }}
     - name: Build and Push Docker Image
       run: |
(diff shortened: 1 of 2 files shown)


jameshughes89 (approved):
Knowing very little about docker really, I can only assume the username change won't break everything. 🎸

twentylemon:
> Knowing very little about docker really, I can only assume the username change won't break everything. guitar

Not much complexity here, it's only a tag change. Instead of build, pushing and running `chippers255/duckbot`, we'll use `duckdynasty/duckbot`. Otherwise, it's identical.

PR #325, opened 2021-09-30
add thx as trigger word for thanking duckbot + only respond once

### LINKED ISSUE #321: Add "thx" as triggers for thanking robot
See https://github.com/duck-dynasty/duckbot/blob/main/duckbot/cogs/robot/thanking_robot.py#L15

Add _thx_ as another "thank you" word option to capture MORE THANKS.

### LINKED ISSUE #322: Only send a single "don't thank robot" message
See https://github.com/duck-dynasty/duckbot/blob/main/duckbot/cogs/robot/thanking_robot.py#L15

Currently, if the message content contains multiple "thanks," duckbot will response twice. Just needs a short circuit.

jameshughes89:
##### Summary 

Closes #321 
Closes #322 

Added "thx" variations on thank you to the `thanks` list of thank yous. This includes "thx duckbot" and "thx duck bot". I considered "thxs" versions too, but I don't think that would be common (but easy to add if we want). 

Also changed the loop checking if a given thank you variation is within the message. Instead of a loop with an if, I use _list comprehension_ (which I typically avoid, but whatever) to make a list of bools and then use `any` to check if any of them are true. I opted for this over a `break` or `return` because I figured there would be a python-ie way to do the check. Also added a test to check that only one reply is sent with a message containing more than one version of thanking.

##### Checklist

### DIFF
--- a/tests/cogs/robot/thanking_robot_test.py
+++ b/tests/cogs/robot/thanking_robot_test.py
@@ -20,6 +20,14 @@ async def test_correct_giving_thanks_message_is_thanks(bot, message, text):
     await clazz.correct_giving_thanks(message)
     message.channel.send.assert_called_once_with(f"I am just a robot.  Do not personify me, {message.author.display_name}")
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("text", ["Thank you DuckBot. thanks duck bot. thx duck bot boy", " tHaNks, DuCK BOt"])
+async def test_correct_number_of_replies_to_very_thankful_messages(bot, message, text):
+    message.content = text
+    clazz = ThankingRobot(bot)
+    await clazz.correct_giving_thanks(message)
+    message.channel.send.assert_called_once_with(f"I am just a robot.  Do not personify me, {message.author.display_name}")
+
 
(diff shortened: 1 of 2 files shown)


tests/cogs/robot/thanking_robot_test.py
```diff
@@ -21,6 +21,15 @@ async def test_correct_giving_thanks_message_is_thanks(bot, message, text):
     message.channel.send.assert_called_once_with(f"I am just a robot.  Do not personify me, {message.author.display_name}")
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("text", ["Thank you DuckBot. thanks duck bot. thx duck bot boy", " tHaNks, DuCK BOt"])
+async def test_correct_number_of_replies_to_very_thankful_messages(bot, message, text):
```
twentylemon, line 26: Excellent test name.

twentylemon (commented):
Looks good. Can you add a case for _thx_? Should just be adding a new param to the `test_correct_giving_thanks_message_is_thanks` test.

twentylemon (approved):

PR #327, opened 2021-09-30
deploy duckbot to aws

twentylemon:
##### Summary 

Righto, a few things to go over. First, moving to AWS more or less to get off of the pi and never deal with arm platforms ever again. Accessibility to the hardware is also much simpler. I can pretty easily add access to the AWS account, and we can set up ssh access to the EC2 host itself.

I'm using [cloudformation](https://aws.amazon.com/cloudformation/) for the infrastructure code. It's basically the same thing as terraform but worse (it's fully yml configured, no code sharing, etc, is possible without things like jinja or CDK).

Infrastructure wise:
* I'm using [ECS](https://aws.amazon.com/ecs/) backed by a single EC2 instance
    * ECS tasks are nearly 1:1 comparable to a docker-compose specification (there's even a tool to convert the two, but it produces results that cost much more money)
    * EC2 has a free tier which is the main reason I chose it over fargate, we'll see what happens after 12 months
    * I jammed the EC2 instance into an autoscaling group to help deal with availability zone outages, or allowing us to replace the host without worrying about bringing a new one back up manually (if the host burns out or whatever)
* For storage, I'm using [EFS](https://aws.amazon.com/efs/) to avoid making changes to the code base
    * as in, we still host our own postgres image on the same instance as the duckbot container
    * I chose EFS over EBS for much simpler persistence; EFS is multi-availability zone, and the lifetime of the is permanent by default, instead of tied to the EC2 instance
* Much of the rest of the cloudformation stack is just networking stuff, or IAM stuff

I ran into trouble mounting EFS to the container at the default postgres storage location. I don't really know why still, but changing the data location fixed all the issues I was having. Volumes for some reason were disappearing once the container stopped.

I've added all the new secrets to the repo already.

And, finally, how it all ties together.
1. `git push` triggers `deploy` action
2. `deploy` builds the `duckdynasty/duckbot` image, and tags it with `latest` and the commit hash
3. create api tokens in [AWS Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html) so ECS can pull from them when it runs duckbot
4. deploy the cloudformation stack using the new image that is tagged with the commit hash
    * ECS recommends using unique image tags per update so it can roll back on failure 
    * the vast majority of the time, the only update to the cloudformation stack will be the docker image to deploy

##### Checklist


jameshughes89:
jesus

twentylemon:
wut

Chippers255 (approved):

PR #330, opened 2021-10-02
correct @mock.patch argument order

twentylemon:
##### Summary 

Title. While working on slash commands stuff, I noticed this. Bottom first for @mock.patch! ([scroll down a bit here](https://docs.python.org/3/library/unittest.mock.html#quick-guide)) Tiny PR to fix it.

##### Checklist

### DIFF
--- a/tests/cogs/announce_day/announce_day_test.py
+++ b/tests/cogs/announce_day/announce_day_test.py
@@ -28,8 +28,8 @@ async def test_cog_unload_cancels_task(bot, dog_photos):
 
 
 @pytest.mark.asyncio
-@mock.patch("random.random", return_value=0.5)
 @mock.patch("random.choice", side_effect=["today", "tomorrow", "yesterday", "{today} {tomorrow} {yesterday}"])
+@mock.patch("random.random", return_value=0.5)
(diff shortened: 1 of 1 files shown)


jameshughes89 (approved):

jameshughes89:
On phone, so searching is annoying, but this note was brought up months ago in a PR comment or something. Sneaky sneaky thing.

jameshughes89:
Looks like #223 & #224

twentylemon:
> On phone, so searching is annoying, but this note was brought up months ago in a PR comment or something. Sneaky sneaky thing.

Yeah. Problem is it applies in the opposite order you'd expect regular function calls to work. `first(second(func))` is the expectation, which would evaluate `second` first, modifying the function by putting the argument, then `first` would operate on that new function, thus making the args appear top first. But they appear bottom first. So like. Yeah.

jameshughes89:
> > On phone, so searching is annoying, but this note was brought up months ago in a PR comment or something. Sneaky sneaky thing.
> 
> Yeah. Problem is it applies in the opposite order you'd expect regular function calls to work. `first(second(func))` is the expectation, which would evaluate `second` first, modifying the function by putting the argument, then `first` would operate on that new function, thus making the args appear top first. But they appear bottom first. So like. Yeah.

I gottcha'.

PR #331, opened 2021-10-02
add slash command options model classes

twentylemon:
##### Summary 

I'm currently working on slash commands (#200) and have working code, but the diff is pretty large. I'm splitting it up into several smaller changes so it'll be easier to review, and it'll hopefully allow for a better overall understanding on how I've chosen to integrate slash commands into discord.py, which doesn't really support it out of the box.

This change adds the slash commands option type, see the [discord docs](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure). These classes are basically taken from my old PR #246. There's more too them than what I've implemented so far. I figure we can add to them if we ever want to in the future, and I've only added what's required and what is actually used by us.

`to_dict()` on the classes is going to be used when constructing the overall slash command to send to discord when creating/updating them.

##### Checklist


duckbot/slash/option.py
```diff
@@ -0,0 +1,67 @@
...
+    def options(self) -> typing.List:
+        return []
+
+    def to_dict(self) -> dict:
+        return {"name": self.name, "description": self.description, "type": self.type, "required": self.required, "options": [x.to_dict() for x in self.options]}
+
+    def __eq__(self, other) -> bool:
+        return self.to_dict() == other.to_dict() if isinstance(other, Option) else False
+
+    def __str__(self) -> str:
```
Chippers255, line 50: I assume based off this code and it's tests that discord.py might be using str/repr of the class? Or is this for your own completeness or usage later?

duckbot/slash/option.py
```diff
@@ -0,0 +1,67 @@
...
+        return str(self.to_dict())
+
+    def __repr__(self) -> str:
```
Chippers255, line 53: Same as above.

duckbot/slash/option.py
twentylemon: Discord.py won't be using the str/repr, they are essentially just for our own debugging, and for better assertion failure messages, that kind of thing. More or less just the thing to do for POJOs.

Discord.py doesn't have much in the ways of actually dealing with slash commands themselves. They have the low level APIs to manage them (create etc), and added things to respond to them, but that's it. The `to_dict()` is the main bridge between this code and discord.py stuff here, that'll be used when creating/updating the slash commands.

Chippers255: POJO or POPO?

Chippers255 (approved):

duckbot/slash/option.py
twentylemon: POPO is PHP. Don't invite that evil into the world.

https://en.wikipedia.org/wiki/Plain_old_Java_object

> For PHP, it is plain old PHP object (POPO).[4][5]

PR #338, opened 2021-10-05
add tests for calculated date special days

twentylemon:
##### Summary 

Adds tests for the special days that are calculated dates. Mother's day for example is the 2nd Sunday of May, or whatever.

Test cases, I pulled from https://www.timeanddate.com/holidays/canada/mother-day

I could also add tests for the rest of them. Stuff like birthdays includes age as a calculation, but I think they're just not required. I don't think the test would catch any logic errors there.

##### Checklist

### DIFF
--- a/tests/cogs/announce_day/special_days_test.py
+++ b/tests/cogs/announce_day/special_days_test.py
@@ -1,4 +1,6 @@
-import datetime
+from datetime import datetime
+
+import pytest
 
 from duckbot.cogs.announce_day.special_days import SpecialDays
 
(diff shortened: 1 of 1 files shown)


tests/cogs/announce_day/special_days_test.py
```diff
@@ -9,4 +11,16 @@ def test_populate_bro_tito_day(bot, guild, emoji):
     emoji.name = "tito"
     guild.name = "Friends Chat"
     clazz = SpecialDays(bot)
-    assert clazz.get_list(datetime.datetime(2020, 5, 7)) == [f"Bro Tito Day {emoji}"]
+    assert clazz.get_list(datetime(2020, 5, 7)) == [f"Bro Tito Day {emoji}"]
+
+
+@pytest.mark.parametrize("date", [datetime(2021, 5, 9), datetime(2022, 5, 8), datetime(2023, 5, 14)])
```
jameshughes89, line 17: Are these hard coded days? I guess there no simple `date.givemesecsundayofmay()`.

twentylemon: That's how it works in the [source file](https://github.com/duck-dynasty/duckbot/blob/main/duckbot/cogs/announce_day/special_days.py#L34), I didn't want to duplicate the logic here to test the logic.

jameshughes89 (approved):

PR #347, opened 2021-10-13
remove `pip --editable` since it is now the default; pin flake8 to <4

twentylemon:
##### Summary 

Had some weird build errors recently, turns out [pip updated recently](https://pip.pypa.io/en/stable/news/#v21-3) to 21.3, which semi-broke our `--editable` build. Not sure what else happened, but it also upgraded our flake8 to 4.0, which breaks `flake8-isort`, so I opted to pin the version to <4.

For testing, I ran all the commands in the readme, as well as running stuff like `python -m duckbot`. The docker image isn't affected as it never used `--editable` anyways.

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -31,7 +31,7 @@ jobs:
         python -m venv --clear venv
         . ./venv/bin/activate
         pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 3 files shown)


jameshughes89 (approved):
Not sure why -e was busting it, but whatevz

PR #354, opened 2021-10-16
restore --editable because pip is confusing

twentylemon:
##### Summary 

Recently removed `--editable` (https://github.com/duck-dynasty/duckbot/pull/347) because docs made it sound like it was the default. Apparently that's not the case, we still need it to avoid installing our package into our own `venv`.

I also changed the github actions cache key for our virtual env used by the tests. The new key will make whatever is currently cached never be accessed, so we'll recreate it. Whatever is there now may conflict with code changes we add, so it could lead to more confusing errors.

James called it out. I DIDN'T LISTEN.

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -23,7 +23,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: venv
-        key: ${{ env.pythonLocation }}-venv-${{ hashFiles('setup.py') }}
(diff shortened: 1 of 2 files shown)


jameshughes89:
I'd love to take credit, but i too, after having a quick read at the docs, didn't think it'd make any difference anyways :/

twentylemon:
moving to my branch to prevent noise for the time being

PR #355, opened 2021-10-16
restore `--editable` because pip is confusing

twentylemon:
##### Summary 

Re-open of https://github.com/duck-dynasty/duckbot/pull/354

Found some context in https://github.com/pypa/pip/issues/7778 -- seems `setup_requires` is deprecated, I guess behaviour for it changed in recent updates. We use it to install nltk corpora during the `pip install`, which means nltk has to be available during the build (ie, the setup). The way to do that now is via `build-system.requires` in `pyproject.toml`. A little annoying, dependencies are split across files, but not sure what else I can do.

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -23,7 +23,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: venv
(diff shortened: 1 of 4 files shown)


jameshughes89 (approved):
I thought i approved this already? Must've not hit submit or something...

PR #362, opened 2021-10-17
swap out `urllib` for `requests`

### LINKED ISSUE #361: swap out `urllib` for `requests`
[requests](https://docs.python-requests.org/en/latest/) is a higher level abstraction replacement for `urllib` ([docs](https://docs.python.org/3.8/library/urllib.request.html) even say prefer `requests`), makes sense to avoid `urllib` and clean up our code/tests a bit.

For testing, we can also use the [responses](https://pypi.org/project/responses/) library to stub behaviour more easily, without writing our own junk as we do now.

twentylemon:
##### Summary 

Fixes #361 

Diff is a little annoying, there's no logic changes really, I only use `requests` instead of `urllib`. For tests, I use `responses` library to help setup and assert calls, which seems to be where the diff just dies. The test suite hasn't changed really, the same smattering of tests is run, just against the new implementation.

##### Checklist


duckbot/cogs/recipe/recipe.py
```diff
@@ -80,6 +74,6 @@ async def __recipe(self, context, search_term):
         await context.send(response)
 
     @commands.command(name="recipe")
-    async def recipe(self, context, *, search_term: str = ""):
+    async def recipe_command(self, context, *, search_term: str = ""):
```
jameshughes89, line 77: sneaky

tests/cogs/games/office_hours_test.py
```diff
@@ -1,20 +1,8 @@
...
-@mock.patch("requests.Response")
-def mock_response(r) -> requests.Response:
-    return r
-
-
-def response(content) -> requests.Response:
-    r = mock_response()
-    r.content = content.encode()
-    return r
+TWITCH_URI = "https://www.twitch.tv/conlabx"
```
jameshughes89, line 5: NIce

jameshughes89 (approved):

PR #368, opened 2021-10-19
add slash command context class

twentylemon:
##### Summary 

I'm currently working on slash commands (#200) and have working code, but the diff is pretty large. I'm splitting it up into several smaller changes so it'll be easier to review, and it'll hopefully allow for a better overall understanding on how I've chosen to integrate slash commands into discord.py, which doesn't really support it out of the box.

This change adds a slash command _context_ class, which is meant to mimic the discord.py [context](https://discordpy.readthedocs.io/en/stable/ext/commands/api.html#context) class, but change some methods (like `send`, etc) to interact with the slash commands APIs instead of message ones.

For tests, I changed the existing `context` fixture to now provide instances of the discord.py context and the new slash context introduced here. Autospec'd mocks allow us to hopefully capture the interface differences between the two contexts before we actually run any real code, since they'll throw when you access non-existent properties. As of this change, every command we have is also being tested as if it is a slash command, we just don't have any slash commands registered yet.

##### Checklist

### DIFF
--- a/duckbot/slash/context.py
+++ b/duckbot/slash/context.py
@@ -1,0 +1,103 @@
+import logging
+from typing import Optional
+
+from discord import Client, Embed, File, Guild, Interaction
+from discord.ext.commands import Command
+from discord.ext.commands.view import StringView
+
+log = logging.getLogger(__name__)
+
+
+def _create_arguments_view(interaction: Interaction, command: Command) -> StringView:
+    # create option string by joining all of the arguments together, so we can delegate to discord.py
+    # note that slash commands can accept spaces in their args, whereas discord.py required "bash style quotes"
+    options = interaction.data.get("options", [])
+    parts = []
+    if options and not options[0].get("value", None):  # a command group
+        if command._discordpy_include_subcommand_name[options[0]["name"]]:
+            parts.append(options[0]["name"])
+        parts = parts + [f'"{x.get("value", "")}"' for x in options[0].get("options", [])]
+    else:
+        parts = parts + [f'"{x.get("value", "")}"' for x in options]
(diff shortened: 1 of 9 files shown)


duckbot/slash/context.py
```diff
@@ -0,0 +1,112 @@
...
+log = logging.getLogger(__name__)
+
+
+def _create_arguments_view(interaction: Interaction, command: Command) -> StringView:
+    # create option string by joining all of the arguments together, so we can delegate to discord.py
+    # note that slash commands can accept spaces in their args, whereas discord.py required "bash style quotes"
+    options = interaction.data.get("options", [])
+    parts = []
+    if options and not options[0].get("value", None):  # a command group
+        if command._discordpy_include_subcommand_name[options[0]["name"]]:
```
twentylemon, line 17: Should note, this attribute will be patched into the command in a follow up PR! This related a command to it's parent, if it has one (eg, `!weather set`; `set` would get connected to `weather`). See [sub-command groups docs](https://discord.com/developers/docs/interactions/application-commands#subcommands-and-subcommand-groups).  
This whole logic is to support `!weather` being able to be invoked without a subcommand, which is not possible in slash commands. We have to send `/weather get`, so this junk is to optionally remove `get` from the argument list.

twentylemon: I may change this to be contained within the `SlashCommand` class (https://github.com/duck-dynasty/duckbot/pull/367), but will leave that for a follow up since it's a little cross cutting.

twentylemon: This is added by #379

Chippers255 (approved):

PR #377, opened 2021-11-04
remove restart_policy conditions in docker compose file

twentylemon:
##### Summary 

Not sure why, but docker started to complain about the `any` for these, even though it's a [valid value](https://docs.docker.com/compose/compose-file/compose-file-v3/#restart_policy). Docs also say `any` is the default, so just removing it fixes.

Error I was getting:
```
Error response from daemon: invalid restart policy 'any'
```
After removing, the stack starts up.

##### Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,6 @@ services:
       mode: replicated
       replicas: 1
       restart_policy:
-        condition: any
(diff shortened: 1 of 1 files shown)


jameshughes89 (approved):
🥑

PR #376, opened 2021-11-02
ensure event exceptions are logged

twentylemon:
##### Summary 

Part of #212, this ensures event errors (like in `on_message` are logged). Tasks are still not handled, you apparently need to register an error handler _per loop_. Trying to think of a clean way to make a global one.

I also refactored the logging module a bit. Just merged everything into one file pretty much. No real logic changes to the moved code. I switched from the `discord` logger to the `duckbot` logger, but they both go to the same log file.

##### Checklist

* [x] this is a source code change
  * [x] run `isort . && black .` in the repository root
  * [x] run `pytest` in the repository root
  * [x] link to issue being fixed using a [_fixes keyword_](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue), if such an issue exists
  * [x] update the wiki documentation if necessary
* [ ] or, this is **not** a source code change

PR #381, opened 2021-11-07
ensure exceptions from task loops are logged

### LINKED ISSUE #212: ensure execptions are logged

twentylemon:
##### Summary 

Fixes #212. Mentioned in #376, task exceptions (ie `@loop`s in discord.py) require an error handler to be registered _per loop_. I don't really want to duplicate that code, so I instead opted for a horrible monkeypatch solution. Essentially, I patch the actual `@loop` with a decorator which adds a default error handler before returning it, and otherwise has the same behaviour as the original `@loop`. So all existing and future `@loop`s automatically get the error handler applied without knowing it.

I also did some minor cleanup. Our logging functions were logging and printing to stderr, but we can save repeating ourselves there by adding two handlers, one for stderr and another for logging to disk. You apparently also don't need to configure all loggers if you configure the root logger, which you can do via [logging.basicConfig](https://docs.python.org/3/library/logging.html#logging.basicConfig).  
Since I was there, I figured I'd also try to use the most appropriate logger when errors occur. I changed the `on_command_error` logger to use a logger from the same module as the owning cog; similarly for tasks (when possible). I couldn't figure out how to do the same for events (like `on_message`), so they are still just using the `duckbot` module logger. This doesn't change much, it's just the logger name that gets printed (like `announce_day` below), which makes it a bit more informative.

Example log; I injected an error into the day of the week announcement:
```
2021-11-06 22:48:13,114:ERROR:duckbot.cogs.announce_day.announce_day: on_hour_loop
Traceback (most recent call last):
File "/opt/venv/lib/python3.8/site-packages/discord/ext/tasks/__init__.py", line 168, in _loop
await self.coro(*args, **kwargs)
File "/duckbot/duckbot/cogs/announce_day/announce_day.py", line 46, in on_hour_loop
raise RuntimeError("ded")
RuntimeError: ded
```

##### Checklist


jameshughes89 (approved):
Monkey patch scares me.

twentylemon:
> Monkey patch scares me.

Your safety is not guaranteed.

PR #394, opened 2021-11-11
🐛 fix `/eightball` args ending with quotes

twentylemon:
##### Summary 

Seems https://github.com/duck-dynasty/duckbot/pull/389 doesn't work perfectly. After some investigation, it seems it's erroneously receiving `"question?"` instead of `question?` as the input string. I construct the args string manually using the options passed into the slash command, but it turns out some internal stuff collided with the command signature just perfectly so to keep unwanted quotes around. Essentially, using a kwarg-only arg told discord.py to consume the rest of the args and jam it into that kwarg -- and for some reason, that's _as is_, even though everything else removes quotes.

Fix is basic enough for now, I just check if there's only one argument to the command, and if so, remove the quotes. I also changed the quotes to be less likely to clash with quotes that people might actually type themselves.  
I fully expect to revisit this as we register more slash commands, but keeping this change focused on fixing `/eightball` for now.

Did some manually testing in the beta `#stonks` channel.

(To face my own face, let it be known that quoting stuff was a last minute addition that I didn't really test.)

##### Checklist

### DIFF
--- a/duckbot/slash/context.py
+++ b/duckbot/slash/context.py
@@ -19,7 +19,9 @@ def _create_arguments_view(interaction: Interaction, command: Command) -> String
         parts = parts + [f'｢{x.get("value", "")}｣' for x in options[0].get("options", [])]
     else:
         parts = parts + [f'｢{x.get("value", "")}｣' for x in options]
-    view = " ".join([x for x in parts])
+
+    # remove wrapping quotes if there's only one arg
+    view = " ".join(parts)[1:-1] if len(parts) == 1 else " ".join(parts)
     log.debug("args string=%s", view)
     return StringView(view)
 
--- a/duckbot/slash/context.py
+++ b/duckbot/slash/context.py
@@ -16,9 +16,9 @@ def _create_arguments_view(interaction: Interaction, command: Command) -> String
     if options and not options[0].get("value", None):  # a command group
         if command._discordpy_include_subcommand_name[options[0]["name"]]:
             parts.append(options[0]["name"])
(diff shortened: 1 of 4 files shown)


duckbot/slash/context.py
```diff
@@ -16,10 +16,12 @@ def _create_arguments_view(interaction: Interaction, command: Command) -> String
     if options and not options[0].get("value", None):  # a command group
         if command._discordpy_include_subcommand_name[options[0]["name"]]:
             parts.append(options[0]["name"])
-        parts = parts + [f'"{x.get("value", "")}"' for x in options[0].get("options", [])]
+        parts = parts + [f'｢{x.get("value", "")}｣' for x in options[0].get("options", [])]
```
Chippers255, line 19: image
｢ & ｣

twentylemon: > I also changed the quotes to be less likely to clash with quotes that people might actually type themselves.

They're valid quotes according to discord.py's [StringView](https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/view.py#L28-L46) class.

Chippers255: Kind of confusing to look at. Could you use a double or single quote doc string in place?
```python
parts = parts + [f'''"{x.get('value', '')}"''' for x in options[0].get("options", [])]
```
Though looking at that, it doesn't seem much better.

twentylemon: The reasoning isn't readability, the options values are user input, so it may include regular ascii quotes. I don't want to deal with escaping all that stuff, so using these freakazoid quotes is probably good enough.

Chippers255 (approved):

PR #400, opened 2021-11-12
:art: use `@dataclass` to clean up boilerplate code

twentylemon:
##### Summary 

Python has [dataclasses](https://docs.python.org/3.8/library/dataclasses.html) built in, which basically add constructors, equals, to string, all that good stuff, for you. I wanted to check them out, so I tested it out on the slash commands model classes added recently. Diff is rather large, but it's mostly removal.

* I took advantage of `__post_init__` that the data class library provides, which runs after `__init__` and can be used to do validation and such. I added validation to slash command and option names, as they can't contain spaces.
* `asdict` will convert a data class into a dictionary directly, so removes some logic. I opted to keep the `to_dict` methods (as opposed to using `asdict` at the call sites) for testing purposes mostly
* `InitVar` is something that's passed into the constructor methods, but not actually stored in the class afterwards (and thus does not show up in `asdict`), I use this in the `SlashCommand` class, there's some contextual data that's only needed during construction (if it's a regular slash command, or a group)

##### Checklist


Chippers255 (commented):
Is there a reason for removing the getters and setters? I imagine they weren't really used much before because python.

twentylemon:
> Is there a reason for removing the getters and setters?

`@dataclass` essentially replaces all of that. We weren't doing anything special in the getters/setters, just wrapping a class field, which is the out-of-the-box behaviour.

Chippers255 (approved):

PR #403, opened 2021-11-13
🚸 mark `/start` and `/stop` as slash commands; support `delete_after` for slash commands

twentylemon:
##### Summary 

This change also implements the `delete_after` parameter of `InteractionContext.send`.

Slash commands get marked by discord as failed if no response is sent to the command within three seconds. `!start` and `!stop` don't send any messages currently, so they get marked as failed, even though `/start` was actually working fine. To avoid confusion there, I changed the commands so they send a message that gets deleted after 30s. Then, for parity, I figured I'd implement `delete_after`. Turned out to be fairly simple, discord.py supports the async delete out of the box, I just had to call it.

Tested new commands manually, and also tested the _followup_ variant for delete after by modifying `/eightball` locally.

##### Checklist


jameshughes89 (approved):
K

PR #406, opened 2021-11-14
:loud_sound: setup sns topic for alarm notifications; notify of stopped containers

### LINKED ISSUE #384: setup cloudwatch alarms
See discussion on https://github.com/duck-dynasty/duckbot/pull/383

CloudWatch alarms are free up to 10 alarms, and basic metrics (5min resolution) are free as well. We can set up some alarms to:
* monitor the running containers (ecs already has health checks, so we may want to monitor too many health check failures?)
* monitor the ec2 instance; alarms can [terminate/reboot the instance](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/UsingAlarmActions.html)

It'd probably be ideal if we can make it so the alarm action notifies the `#duckbutt` channel. Quick google says this might be doable as [alarm -> sns -> lambda](https://towardsdatascience.com/discord-notification-using-cloudwatch-alarms-sns-and-aws-lambda-71393861699f), but lambda isn't quite free. Given the expected low volume though, should still cost [next to nothing](https://aws.amazon.com/lambda/pricing/).

twentylemon:
##### Summary 

Note: maybe do https://github.com/duck-dynasty/duckbot/pull/423 first

I guess fixes https://github.com/duck-dynasty/duckbot/issues/384, though we can add more alarms. If you have better ideas for what to alarm on, let me know. 

Spent a while looking into all the different alarming options:
* ec2 instance alarms; would allow termination of a dead instance (say, CPU redline), but alarm is per-instance, and would have to be recreated when the alarm actually triggered or if the instance is otherwise replaced
* autoscaling scaling alarms; would allow de-scaling and re-scaling, but requires chaining alarms together (say CPU redline -> downscale -> no instances -> upscale) which I didn't really like
* ecs container alarms; what I ended up at

I figured:
* ec2 already does health checks, and I don't really care if the autoscaling group replaces a host; there's nothing we can really do to react to such an alarm
* we've configured health checks on the ecs containers, so they'll be restarted if health checks fail (or if the containers crash)
* autoscaling alarms don't really matter if ecs is already going to restart tasks

So, being notified of unintentionally stopped ecs containers seemed like enough, at least for starters. See [tutorial](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_cwet2.html) I mostly followed.

Note this doesn't actually add subscriptions, we can do that in follow up commits. ~~It'd be specifying the `AlarmSubscription*` cloudformation parameters, which would likely be github secrets.~~ We'll need a new webhook to publish to.

Tested manually by creating the topic, rule and subscribing myself. I then ssh'd into the ec2 host and did `docker kill`. ~~Received a single email for the dead instance.~~ See `#duckbutt` channel in the beta server.

##### Checklist

  * [ ] run `format` in the repository root


twentylemon:
Seems lambda's free tier is more than enough to cover alarms. Moving to draft, I'll fiddle with publishing to a discord webhook instead.

twentylemon:
> Seems lambda's free tier is more than enough to cover alarms. Moving to draft, I'll fiddle with publishing to a discord webhook instead.

Now it's all lambda. Neat.

twentylemon:
Revisiting due to the merge, I'm not sure it's a great alarm. Stopped containers still get restarted by ECS, so it's still an alarm we can't really action on.

Will re-think in the future. Just closing.

PR #422, opened 2021-11-20
👷 remove coverage from default `pytest` run for local speedup

twentylemon:
##### Summary

Speeding up unit tests actions a bit, was just looking around at different stuff available in `pytest`. This was originally splitting up tests across multiple actions workers, but I figured it was needlessly complicating. Instead, this change:

* removes coverage from the default `pytest` invocation (coverage adds a fair bit of slowdown, like 50-80% longer runtime)
    * can always run coverage by adding args; updated the readme
* `release` no longer fails on low coverage, instead letting the codecov comment/check to report on that
* renamed `release` to `test`, and made a dummy `release` job which always succeeds when `test` and `sanity` succeed
    * this was originally to deal with having matrix jobs be part of branch protection, now it's just a clean way to add more job requirements without having to change the branch protection rule (eg, if we add service tests some day)
    * updated the branch protection rules accordingly, so `release` is the only action required, and `codecov/project` is required (so, 85% code coverage, same as before)

##### Checklist


jameshughes89 (approved):
Sweeeeet/

PR #426, opened 2021-11-25
👷 add stale pull request manager

twentylemon:
##### Summary

Recently marked some stuff as `needs reply`, which sparked a google search to see if something exists to manage that crap automatically. Turns out github has their own [action](https://github.com/actions/stale). It can mark things as stale, then close stale things.

I think I set it up so that we have to apply `needs reply` before it'll actually do anything.  
After a month of that, it'll attach `stale`, then another month it'll get closed. I chose a long time frames since it's a hobby project.  
It should only act on pull requests, not issues.

##### Checklist


jameshughes89 (approved):

.github/workflows/close-stale-pulls.yml
```diff
@@ -0,0 +1,27 @@
+name: Stale Issues/Pull Requests
+
+permissions:
+  issues: write
+  pull-requests: write
+
+on:
+  schedule:
+  - cron: '7 13 * * MON'
```
jameshughes89, line 9: is there method to the madness of 7 13?

twentylemon: Yup. Ideally for scheduled jobs you want to add jitter, that is, some randomness to when it actually starts. Reason being relates to hardware -- most everyone is going to run stuff at the top of the hour, there's a big race for hardware, potentially even causing some scaling failures. Jitter distributes the job fairly evenly across a time period, so the other services have time to scale up or don't even need to scale up.  
You can't add jitter cron, so essentially just picking a time that people are not likely to pick.

13 UTC is morning our time, so it also not going to produce too much noise like overnight or whatever.

PR #427, opened 2021-11-25
👷 add local python setup composite action

twentylemon:
##### Summary

[Composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) are a way to re-use a set of actions, and you can actually reference [local actions](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#referencing-an-action-in-the-same-repository-where-a-workflow-file-uses-the-action). Was looking into this due to the duplication created in the actions in #423. I probably won't change #423 if this gets merged first, and will issue another pull request for some cleanup after.

##### Checklist


jameshughes89 (approved):
🤞

PR #428, opened 2021-11-25
🚸 only register global slash commands in prod

### LINKED ISSUE #395: slash commands are duplicated in guild
Slash commands have two scopes, global commands and guild specific commands. We currently [register both kinds with discord](https://github.com/duck-dynasty/duckbot/blob/main/duckbot/slash/slash_command_handler.py#L26-L33). This leads to showing duplicate slash commands in the guild, as the discord client shows both. I haven't seen any way to hide them ([docs](https://discord.com/developers/docs/interactions/application-commands)).

Some things to consider:
* guild commands cannot be used in a direct message, or otherwise outside of the guild
* global commands can take up to an hour to appear in the discord client, whereas guild commands are not cached and are available immediately

Guild commands are useful while testing since you don't have to wait. Global commands are more generally useful in prod, but take a while to show up. Maybe a decent solution is to only register global commands in prod, and only register guild commands otherwise? The `!` commands are available instantly in prod, so it's at least a workaround while waiting for the available slash commands to update. Or maybe just always using guild commands, and never being able to use slash commands in private messages -- though `!` commands could still be used.

twentylemon:
##### Summary

Fixes https://github.com/duck-dynasty/duckbot/issues/395

Decided to go with only registering global slash commands in prod, but registering both in non-prod. This means slash commands will take ~1h to actually show up in prod, but old style `!` commands are still available instantly.

For manual testing, I set `STAGE=prod` environment variable, then ran duckbot. Saw log:
```
2021-11-25 18:09:55,713:INFO:duckbot.slash.slash_command_handler: registering global slash commands=[{SNIP}]  
2021-11-25 18:09:56,634:INFO:duckbot.slash.slash_command_handler: registering slash commands in guild={redacted} commands=[]
```

Note, ~~this sort of conflicts with https://github.com/duck-dynasty/duckbot/pull/423. I went ahead and added the `STAGE` environment variable to that PR, even though it's not really used yet.~~ https://github.com/duck-dynasty/duckbot/pull/423 has already added the `STAGE` environment variable to the prod deployment.

##### Checklist


jameshughes89 (approved):
🚅

PR #446, opened 2021-12-01
:bug: :rocket: move duckbot image tag to cdk context variable

twentylemon:
##### Summary

CDK seems to have a [bug](https://github.com/aws/aws-cdk/issues/8734) where new parameter values aren't picked up if there's no other stack changes. This means the docker image for duckbot is never actually updated. There's a workaround for now to use `--force`, which I am using. Though, seems the best method is to actually use a context variable instead, so it's available at synthesis time instead of deployment time. I figure let's do both, screw it.

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -198,7 +198,7 @@ jobs:
         echo "::endgroup::"
         echo "::group::cdk deploy"
         cdk deploy \
-          --all \
+          --force --all \
           --context duckbot_image=${{ steps.image.outputs.commit }} \
           --context write_secrets=true \
           --require-approval never \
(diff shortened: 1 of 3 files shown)


.github/workflows/python-package.yml
```diff
@@ -191,15 +191,15 @@ jobs:
...
         echo "::endgroup::"
         echo "::group::cdk diff"
-        cdk diff --parameter DuckBotImage=${{ steps.image.outputs.commit }} || true
+        cdk diff --context duckbot_image=${{ steps.image.outputs.commit }} || true
         echo "::endgroup::"
         echo "::group::cdk deploy"
         cdk deploy \
-          --all \
-          --parameter DuckBotImage=${{ steps.image.outputs.commit }} \
+          --force --all \
```
jameshughes89, line 201: :muscle:

jameshughes89 (approved):

PR #495, opened 2022-02-01
🚨 change pytest-asyncio to 'auto' mode instead of deprecated 'legacy'

### LINKED ISSUE #494: address `pytest-asyncio` deprecation warnings
See warnings on [test run](https://github.com/duck-dynasty/duckbot/runs/5021993377?check_suite_focus=true)

twentylemon:
##### Summary

Fixes https://github.com/duck-dynasty/duckbot/issues/494. Big diff, but it's mostly the removal of now unused decorators.

pytest-asyncio recommends using `auto` mode, which is what I set it to. On their [readme](https://github.com/pytest-dev/pytest-asyncio#modes), they mention you also don't need to use `@pytest.mark.asyncio` any more, which is nice, saves some noise in all the tests. I've done through and removed all those as well.

I also fixed another deprecation warning related to upgrades, `responses` no longer uses the `match_querystring` parameter, so I just removed it. See their [readme](https://github.com/getsentry/responses#query-parameters-matcher).

##### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 200
-target-version = ['py38']
+target-version = ["py38"]
 
 [tool.isort]
 profile = "black"
@@ -17,3 +17,4 @@ addopts = """\
     --mdformat \
     -n auto \
     --blockage"""
+asyncio_mode = "auto"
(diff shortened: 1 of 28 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 200
-target-version = ['py38']
+target-version = ["py38"]
```
jameshughes89, line 7: Nice

jameshughes89 (approved):
🚀

PR #565, opened 2022-08-19
💚 enable "legacy-editable" mode for setuptools

twentylemon:
##### Summary

Setuptools updated recently and changed how editable mode works, which we use for development, see [docs](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior). Tested locally by running the updated commands in the readme + pytest.

The docker build was not affected, since it does not use an editable install. I also tested this locally.

##### Checklist


jameshughes89 (approved):
:shrug:

PR #581, opened 2022-09-11
🚀 change ec2 instance type to t3.nano

twentylemon:
##### Summary

Fixes #357

Don't merge yet -- waiting until Sept 30th to squeeze as much of the free tier as possible.

Context for the image sizing is in the issue.

**Update**; the new host only has 500mb of memory, but the ECS task will still try to allociate ~1gb. Reducing that to the ~0.5gb.

`cdk diff` produces:
```
Resources
[~] AWS::ECS::TaskDefinition TaskDefinition TaskDefinitionB36D86D9 replace
 ├─ [~] ContainerDefinitions (requires replacement)
 │   └─ @@ -70,7 +70,7 @@
 │      [ ]   "StartPeriod": 30,
 │      [ ]   "Timeout": 10
 │      [ ] },
 │      [-] "Image": "duckdynasty/duckbot:9f24417",
 │      [+] "Image": "duckdynasty/duckbot:latest",
 │      [ ] "Links": [
 │      [ ]   "postgres"
 │      [ ] ],
 └─ [~] Memory (requires replacement)
     ├─ [-] 960
     └─ [+] 475
[~] AWS::AutoScaling::LaunchConfiguration AutoScalingGroup/LaunchConfig AutoScalingGroupLaunchConfigDEEB160C replace
 └─ [~] InstanceType (requires replacement)
     ├─ [-] t2.micro
     └─ [+] t3.nano
```
The image difference is also expected, since I did not specify a commit to deploy, so the script picks `latest`.

##### Checklist

- [x] this is a source code change
  - [x] run `format` in the repository root
  - [x] run `pytest` in the repository root
  - [x] link to issue being fixed using a [_fixes keyword_](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue), if such an issue exists
  - [x] update the wiki documentation if necessary
- [ ] or, this is **not** a source code change

PR #582, opened 2022-09-19
🚀 reduce ebs storage size to 10gb

### LINKED ISSUE #405: reduce default ebs size to reduce costs
EBS volumes are priced per GB-month, and the default ECS instances attached 30Gb by default. 30Gb is exactly the free tier, which expires after the year. We can likely get away with reducing it a bit (still need space for the docker images, OS, logs, etc) to reduce the cost a bit.

* https://aws.amazon.com/premiumsupport/knowledge-center/cloudformation-root-volume-property/
    * describes changing an `AWS::ECS::Instance`, but should also apply to `AWS::AutoScaling::LaunchConfiguration`
* https://aws.amazon.com/ebs/pricing/
    * at time of writing, `gp3` is the cheapest storage, but you pay for IO; `gp2` you only pay for storage 
    * unfortunately, hard drives are [not bootable](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html), and can't be used

twentylemon:
##### Summary

Fixes #405 ; Similar to #581, don't merge until ~sept 30, as we can still take advantage of the free tier for a bit longer.

Note, unrelated to this, I aliased names like `aws_whatever` to `whatever`. It adds some noise to the diff.

By default, ec2 provisions 30gb storage, but we don't need anywhere near that much storage, and we have to pay for the storage provisioned. This change should override the default storage (`/dev/xvda` ; see link in issue for details) with a smaller one. I also changed the type of storage to `gp3` instead of the default `gp2` (_general purpose generation n_). Third generation is a bit cheaper (it allows for provisioning more IO throughout on the drive, but whatever comes with it by default is more than sufficient).

`cdk diff` produces:

```
Resources
[~] AWS::ECS::TaskDefinition TaskDefinition TaskDefinitionB36D86D9 replace
 └─ [~] ContainerDefinitions (requires replacement)
     └─ @@ -70,7 +70,7 @@
        [ ]   "StartPeriod": 30,
        [ ]   "Timeout": 10
        [ ] },
        [-] "Image": "duckdynasty/duckbot:9f24417",
        [+] "Image": "duckdynasty/duckbot:latest",
        [ ] "Links": [
        [ ]   "postgres"
        [ ] ],
[~] AWS::AutoScaling::LaunchConfiguration AutoScalingGroup/LaunchConfig AutoScalingGroupLaunchConfigDEEB160C replace
 └─ [+] BlockDeviceMappings (requires replacement)
     └─ [{"DeviceName":"/dev/xvda","Ebs":{"VolumeSize":10,"VolumeType":"gp3"}}]
```
The image difference is also expected, since I did not specify a commit to deploy, so the script picks `latest`.

##### Checklist

### DIFF
--- a/.aws/duckdeploy/stack.py
+++ b/.aws/duckdeploy/stack.py
@@ -81,6 +81,7 @@ class DuckBotStack(core.Stack):
             desired_capacity=1,
             machine_image=aws_ecs.EcsOptimizedImage.amazon_linux2(),
             instance_type=aws_ec2.InstanceType("t2.micro"),
+            block_devices=(aws_ec2.BlockDevice(device_name="/dev/xvda", volume=aws_ec2.BlockDeviceVolume.ebs(volume_size=15, volume_type=aws_ec2.EbsDeviceVolumeType.GP3))),
             key_name="duckbot",  # needs to be created manually
             instance_monitoring=aws_autoscaling.Monitoring.BASIC,
             vpc=vpc,
--- a/.aws/duckdeploy/stack.py
+++ b/.aws/duckdeploy/stack.py
@@ -1,6 +1,12 @@
 from typing import List
 
-from aws_cdk import aws_autoscaling, aws_ec2, aws_ecs, aws_efs, aws_logs, aws_ssm, core
+from aws_cdk import aws_autoscaling as autoscaling
+from aws_cdk import aws_ec2 as ec2
+from aws_cdk import aws_ecs as ecs
+from aws_cdk import aws_efs as efs
+from aws_cdk import aws_logs as logs
+from aws_cdk import aws_ssm as ssm
+from aws_cdk import core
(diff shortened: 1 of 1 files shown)


jameshughes89:
`ecs` == `gp3`?

twentylemon:
> `ecs` == `gp3`?

AWS has a crapload of acronyms; tldr for this change is:
* ecs = elastic container service; this basically lets us say "deploy **this** docker image"
* ec2 = elastic compute cloud; this is the host ecs is deploying to
* ebs = elastic block store; ec2 hosts need storage, this is what they use
* `gpX` = general purpose generation X; ebs introduced `gp3` a while back as new generation SSDs, which are a bit cheaper than `gp2`

jameshughes89 (approved):

PR #602, opened 2022-10-06
🚀 change spot interruption behaviour to terminate ec2 instance

twentylemon:
##### Summary

Title. This effectively also allows the ebs volume attached to the instance to be removed when the instance dies, so we don't end up paying for it. There's nothing on the volume ever worth keeping. 

##### Checklist


jameshughes89 (approved):
neato

PR #624, opened 2022-11-14
👷 address github actions set-output deprecation

twentylemon:
##### Summary

Our actions have a fair bit of warnings on them, [example run](https://github.com/duck-dynasty/duckbot/actions/runs/3463322710). `set-output` is being deprecated for whatever reason.

There are several other warnings, but they seem to be from other actions we depend on, like `setup-python`. The `deploy` step for example:
```
Please update the following actions to use Node.js 16: actions/checkout, docker/setup-buildx-action, actions/cache, docker/login-action, actions/checkout, actions/checkout, actions/setup-python, actions/cache, actions/setup-node, aws-actions/configure-aws-credentials, aws-actions/configure-aws-credentials, actions/cache, actions/setup-python, actions/checkout, actions/setup-node, actions/checkout, docker/login-action, actions/cache, docker/setup-buildx-action, actions/checkout
```
~~Dependabot should [update these automatically](https://github.com/duck-dynasty/duckbot/blob/main/.github/dependabot.yml#L9). I imagine these are also what all the `save-state` deprecation warnings are from too -- we don't use that command.~~  
It seems dependabot does not update these nested actions. I've updated the ones that have updates.

See [blog post](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) about the deprecation.

##### Checklist

### DIFF
--- a/.github/actions/build-push-docker-image/action.yml
+++ b/.github/actions/build-push-docker-image/action.yml
@@ -27,17 +27,17 @@ runs:
       echo "commit=duckdynasty/duckbot:$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
 
   - name: Set up Docker Buildx
-    uses: docker/setup-buildx-action@v1
+    uses: docker/setup-buildx-action@v2
 
   - name: Set up Docker Build Cache
-    uses: actions/cache@v2
+    uses: actions/cache@v3
     with:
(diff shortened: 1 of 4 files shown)


jameshughes89 (approved):
The numbers are bigger, therefore they must be better

PR #641, opened 2022-12-08
:bug: Special Days --- Fix duckbot's inception day seconds calculation

### LINKED ISSUE #640: :bug: Special day: Duckbot inception day --- Wrong alive time
[Within `announce_day/special_days.py`, duckbot says it's alive time in seconds incorrectly. ](https://github.com/duck-dynasty/duckbot/blob/2cae4ecc4402812cb9067ae6e2f10d538bb250c0/duckbot/cogs/announce_day/special_days.py#L52)

Right now it uses the `seconds` attribute, but it [turns out the `seconds` attribute has a range of 0 --- 86399s which is 1 day. ](https://docs.python.org/3/library/datetime.html#timedelta-objects)

[Instead, the `total_seconds()` method would do the trick as it returns the total number of seconds from all resolutions of time in the datetime. ](https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds)

In other words

Change 

```python
(now()-datetime(2020, 12, 3, 10, 39, tzinfo=timezone())).seconds
```
to 

```python
(now()-datetime(2020, 12, 3, 10, 39, tzinfo=timezone())).total_seconds()
```

jameshughes89:
##### Summary

Closes #640 

- Fix the duckbot total seconds alive calculation by using the `totalSeconds()` method 
- Update the unittest to have hardcoded values. 

**NOTE** I kinda hate how format formatted the stuff in the parameterized test annotation 

##### Checklist

### DIFF
--- a/duckbot/cogs/announce_day/special_days.py
+++ b/duckbot/cogs/announce_day/special_days.py
@@ -49,5 +49,5 @@ class SpecialDays(holidays.Canada):
         self[date(year, 11, 14)] = f"Male Kelly's birthday... Maybe... idk. He's around {year - 1989} years old now."
         self[date(year, 11, 1) + rd(weekday=FR(+4))] = "Black Friday. I hope I can get some new socks"
         self[date(year, 12, 2)] = f"Female Kelly's Birthday. She's {year-1989} years old"
-        self[date(year, 12, 3)] = f"DuckBot's Inception Day. I'm about {(now()-datetime(2020, 12, 3, 10, 39, tzinfo=timezone())).seconds}s old"
+        self[date(year, 12, 3)] = f"DuckBot's Inception Day. I'm about {(now()-datetime(2020, 12, 3, 10, 39, tzinfo=timezone())).total_seconds()}s old"
         self[date(year, 12, 5)] = f"Taras' Birthday. He's {year-1989} years old"
--- a/tests/cogs/announce_day/special_days_test.py
+++ b/tests/cogs/announce_day/special_days_test.py
@@ -35,10 +35,11 @@ def test_populate_black_friday(bot, date):
 
 
 @mock.patch("duckbot.cogs.announce_day.special_days.now")
-@pytest.mark.parametrize("time", [datetime(2021, 12, 3, 11, tzinfo=timezone()), datetime(2022, 12, 3, 7, tzinfo=timezone())])
-def test_populate_duckbot_day(now, bot, time):
-    initial_commit_datetime = datetime(2020, 12, 3, 10, 39, tzinfo=timezone())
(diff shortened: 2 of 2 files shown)


jameshughes89:
Marked as draft as there may be an issue with the `now` calculation, as brought up by @twentylemon in #640

duckbot/cogs/announce_day/special_days.py
```diff
@@ -49,5 +49,5 @@ def _populate(self, year):
         self[date(year, 11, 14)] = f"Male Kelly's birthday... Maybe... idk. He's around {year - 1989} years old now."
         self[date(year, 11, 1) + rd(weekday=FR(+4))] = "Black Friday. I hope I can get some new socks"
         self[date(year, 12, 2)] = f"Female Kelly's Birthday. She's {year-1989} years old"
-        self[date(year, 12, 3)] = f"DuckBot's Inception Day. I'm about {(now()-datetime(2020, 12, 3, 10, 39, tzinfo=timezone())).seconds}s old"
+        self[date(year, 12, 3)] = f"DuckBot's Inception Day. I'm about {(now()-datetime(2020, 12, 3, 10, 39, tzinfo=timezone())).total_seconds()}s old"
```
twentylemon, line 52: So what I mean here for `year`, this method mutates this class, adding dates to some internal map like `{"2020-01-01": ["New Year's"]}`. This method is lazily called when you ask for a date in a particular year.

So let's imagine a typical use case.

```python
days = holidays.get(today_in_2022)
# ... the next day
days = holidays.get(tomorrow_in_2022)
```
The first call ends up populating the special days for `2022`. DuckBot's inception day gets some value like `"I'm n seconds old"`. In the second call, there's no need to populate the days again, so it is not called. Say tomorrow is DuckBot's special day, we end up saying `"I'm n seconds old"`, but it's actually `n+1 day` seconds. `now()` is not recalculated.

DuckBot _currently_ turns off every day, but already was not the case in the past, so we can't rely on `_populate` being called daily. We also can't force call it, since `self[date] = x` actually appends to the internal list of days the class keeps. `now()` needs to somehow be based on the `year`. We could just say `datetime(year, 12, 3, 7, 0)`, but that also loses accuracy. Not sure what the best approach is here.

jameshughes89: Perhaps this is a larger issue that should be it's own, separate issue.

jameshughes89: 👀

twentylemon: Sure. Make the issue and I'm game.

jameshughes89:
Marked as ready now that #641 exists.

twentylemon (approved):

PR #645, opened 2022-12-21
👷 use cache property in `setup-*` actions instead of manually caching

twentylemon:
##### Summary

The `setup-python` and `setup-node` actions have been updated at some point to include caching behaviours, so I figure let's use them instead of rolling our own caching.

~~Draft while I fiddle with the actions.~~

* [first run](https://github.com/duck-dynasty/duckbot/actions/runs/3752316141/jobs/6374328061) downloaded dependencies
* [second run](https://github.com/duck-dynasty/duckbot/actions/runs/3752335442/jobs/6374370438) used cached dependencies

It seems slower than what we had before. But whatev.

##### Checklist


twentylemon:
Discarding. It provides no real speed up since we download `nltk` data during the install, and that is not cached by the action out of the box.

jameshughes89:
image

jameshughes89:
awhhhh. gif didn't play

PR #668, opened 2023-01-08
🐛 wrap `/define` in typing

twentylemon:
##### Summary

I used `/define` and got no response recently, may be because discord demands any response from a slash command within like 3 seconds iirc. Sending "thinking" is a valid response we use in a lot of other commands, so adding it to `/define` as well. 

##### Checklist


twentylemon:
Coverage drop is expected, the `typing()` needs to be added to a command method, which are not invoked directly in the tests.

jameshughes89 (approved):
_typing_

PR #671, opened 2023-01-15
❓ Wellness Command --- Add badges for making healthy choices/doing healthy things 

jameshughes89:
##### Summary

- Adds a `\wellness` command to add "badges" (reactions) to your messages for the day. 
- Current badges include:
    - 🛏️ for going to bed the night before at a good hour
    - 🏃 for exercising 
    - 🥗 for choosing a healthy meal 
    - 💩 since pooping is healthy 
    - 🚰 for being mindful of hydration 
    
- It is trivial to add more badges (like meditation 🧘, taking time to stand while at work , or maybe posture checks🧍‍♂️)
- When setting a badge, duckbot will reply with an announcement that the person earned the badge
- If trying to set a badge that doesn't exist, duckbot warns that it's not a badge and politely suggests you send a PR about it, brother!
- Earning badges is honour system and in the eye of the beholder, though, badge earners should expect to explain themselves if sus. 
- Badges are cleared daily at hour `0`, but I imagine the daily shutdown/restart would do the trick anyways. 

The motivation for this command comes from just having a fun way to show off a good choice. The _funnness_ may also encourage people to just drink more water, for example. It may also add positive peer pressure to, for example, eat healthier if you see everyone else making good choices. 

As of now, no unit tests are written, but some manual testing was done. The reason this PR is incomplete is that when manually testing, a concern was raised about the amount of "noise"/how "busy" the messages got. See the below image for an example. 

image

To be fair, it's nothing like the dank channel, but there is still a concern about the noise that's added to just the general conversation. 

Regardless, I am looking for suggestions on this --- Do we think it's fine with the reaction badges? Or are the reactions indeed _too much_?

Potential ideas:
- Can discord roles work here?
- Only have reactions in certain channel(s)? 
- Have reactions only pop up periodically somehow?
- Have reactions only reply to last message in a string of messages by the same person?
- ...
- ...
- ...

I'll probably hold off on further changes on the PR until there are some ideas flowing. 

##### Checklist

**NOTE:** This PR is incomplete as it and is only here to seek feedback. Tests are currently not written and the wiki has not been updated. 

- [X] this is a source code change
  - [X] run `format` in the repository root


duckbot/cogs/wellness/wellness.py
```diff
@@ -0,0 +1,52 @@
...
+    @commands.Cog.listener("on_message")
+    async def react_with_badge(self, message):
+        if message.author.id in self._badges:
+            await asyncio.wait([message.add_reaction(badge_symbols[badge]) for badge in self._badges[message.author.id]])
+
+    def cog_unload(self):
+        self.on_hour_loop.cancel()
+
+    def should_clear_badges(self):
+        return now().hour == 0
```
jameshughes89, line 44: Currently busted --- needs the `datetime` from `announce_day`

duckbot/cogs/wellness/wellness.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @tasks.loop(hours=1.0)
+    async def on_hour_loop(self):
+        await self.clear_badges()
+
+    async def clear_badges(self):
+        if self.should_clear_badges():
+            self._badges = {}
```
jameshughes89, line 52: this is apparently faster than `self._badges.clear()`

twentylemon: Reassignment is also thread safe, so basically always the preferred option.

duckbot/cogs/wellness/wellness.py
```diff
@@ -0,0 +1,52 @@
...
+            await asyncio.wait([message.add_reaction(badge_symbols[badge]) for badge in self._badges[message.author.id]])
+
+    def cog_unload(self):
+        self.on_hour_loop.cancel()
+
+    def should_clear_badges(self):
+        return now().hour == 0
+
+    @tasks.loop(hours=1.0)
+    async def on_hour_loop(self):
```
twentylemon, line 47: I'd prefer no loop I think. Instead of storing only badges, you could also store the time at which the user earned the badge. In `on_message`, you'd have to check if it's recent enough to warrant a reaction. More computationally expensive, but quite a bit simpler and far less state to keep track of. It's maybe 30 datetime checks per message, if everyone submits a badge in our server. /shrug

This could also help with the noise problem. We'd have a lot finer control over how long a badge gets published for. Something like 5 minutes may make sense to reduce noise? Assuming we stick with the reactions approach.

twentylemon:
I kinda like roles, but I'm not sure they offer enough to overcome their limitations. They're quite a bit more subtle, I'd just get like a blue or brown nameplate -- a big plus imo. The downsides though are you can only ever have one colour for your nameplate, and it'd introduce a lot of complexity with role management. If the bot crashes or something, another instance would have to know to _remove_ a role after some time, we can't just rely on an in memory check.

Posting like once per 30s for x minutes may make sense, and does not introduce a whole lot more state management. If we keep the timestamp of when someone acquired a badge, it's not much more to also keep the timestamp of when that badge was last published.

jameshughes89:
@Chippers255, you have any thoughts?

twentylemon:
@Chippers255 we are nothing without your input.

twentylemon:
@Chippers255 :(((((

jameshughes89:
> @Chippers255 :(((((

I'd have also @ed Dave, but I can't, so I didn't.

Chippers255:
The PR description could be improved. I am not sure what your problem is? Is it tagging emojis on every post by that person after they use a wellness command? Could you not have a dedicated leaderboard channel that posts everyone's wellness badges once per hour or so?

jameshughes89:
Closing since as of now I don't think we've got a _good_ way to show the badges yet. Note, the closing does not mean that general idea is dead --- perhaps it just needs some time to brew.

PR #822, opened 2024-01-01
:arrow_up: bump python from 3.8 to 3.10

twentylemon:
##### Summary

3.8 is [dying soon](https://devguide.python.org/versions/), and a lot of dependency upgrades are leaving it behind. This bumps python a few versions to ~~3.12~~ 3.10 -- apparently 3.11+ fails to run on github actions runners. I don't want to deal with figuring out why.

`flake8` doesn't work with pytest any more, so I just removed that check entirely. I added it to the `format` script and to github actions instead.

Most of the change was uneventful, but for some reason the `AnnounceDay` and some other tests went haywire. The patches stopped working, and I don't really know why, there were no changes to that module in the python updates. There's some workarounds there essentially, flipping from relative to absolute imports, and injecting data instead of mocking it.  
For a lot of these, the errors were not consistent. Like 1/10 test runs would fail, and they'd never fail when you ran just the one module. Not really sure what was going on there.

Lastly, `asyncio.wait` was changed, `asyncio.gather(*list)` is the same behaviour.

##### Checklist

### DIFF
--- a/.github/actions/setup-python-venv/action.yml
+++ b/.github/actions/setup-python-venv/action.yml
@@ -23,7 +23,7 @@ runs:
   - name: Set up Python
     uses: actions/setup-python@v4
     with:
-      python-version: '3.11'
+      python-version: '3.10'
 
   - name: Create path Output
     id: path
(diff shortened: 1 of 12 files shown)


jameshughes89 (approved):
I kept having issues with `AnnounceDay` on my work comp. I'd investigated if it's related to the issue here if I cared more.

PR #910, opened 2024-09-01
✨ satisfactory solver commands

twentylemon:
##### Summary

Threw this together in scala yesterday and figured what the heck, a duck's gotta work. This introduces a swathe of new commands under `/satisfy` to manipulate a factory construction, then eventually optimize it.

The optimizer is prety naive, but works pretty well. It's naive in the sense that there's no real cost function, so it will do maximization in a dumb or unexpected way. For example I could say "maximize plastic and rubber" for an oil production chain, and it'll happily pick one or the other since they produce in the same amounts. More expected would be it keeps them roughly in line with each other, rather than prioritizing one. It will however still find optimal if it's feasible to solve. It's also very speedy on my machine; these types of problems are actually easy for computers to solve despite being NP-complete/hard.

There's no tests. And I left todo notes in a comment instead of creating an issue. Sue me.

##### Checklist

### DIFF
--- a/duckbot/cogs/games/__init__.py
+++ b/duckbot/cogs/games/__init__.py
@@ -3,6 +3,7 @@ from .coin_flip import CoinFlip
 from .dice import Dice
 from .discord_activity import DiscordActivity
 from .office_hours import OfficeHours
+from .satisfy import setup as satisfy_setup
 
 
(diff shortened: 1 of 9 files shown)


jameshughes89 (approved):
120% for this.

PR #931, opened 2024-09-09
✨ prefer not leaving excess non-sinkable items

### LINKED ISSUE #917: make `/satisfy` try to prefer not leaving non-sinkable items
Example:
image

HeavyOilResidue can't be sunk, so you need to do another solve to figure out how to get rid of it. Eg, in this case you could `/satisfy output PolymerResin 200` and `/satisfy maximize AwesomeTicketPoints` to get it to turn the goo into coke or something then sink it, but it's extra steps. And it'd require clearing the factory first to remove the existing maximizes and turn them into outputs, else it may find a weird solution that balances out other resources.

This would need to be done with changes to the cost function, which is currently only maximizing the sum of items provided in `maximize`. Items that aren't sinkable would have to cost a negative amount (objective is maximized) so it'd prefer more buildings to be able to sink stuff.

twentylemon:
##### Summary

Fixes https://github.com/duck-dynasty/duckbot/issues/917

Had to re-jig some of the weightings in the objective function. Not sure why but it started to prefer 0.8 250% clocked constructors for the test `test_optimize_two_step_many_sloop_and_power_shards_returns_chain`. 0.8 @ 250% is literally a single 200% machine, so it should have tried to remove a power shard. Increasing the power shard usage cost made it work again. Optimizers.

##### Checklist

### DIFF
--- a/duckbot/cogs/games/satisfy/item.py
+++ b/duckbot/cogs/games/satisfy/item.py
@@ -27,6 +27,7 @@ class Item(Enum):
     PackagedWater = (auto(), Form.Solid, 130)
     Plastic = (auto(), Form.Solid, 75)
     Rubber = (auto(), Form.Solid, 60)
+    PetroleumCoke = (auto(), Form.Solid, 20)
 
     AwesomeTicketPoints = (auto(), Form.Aux, 0)
     MwPower = (auto(), Form.Aux, 0)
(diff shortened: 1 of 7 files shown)


jameshughes89 (approved):

twentylemon:
Not sure why sonar didn't run. I'ma just yolo it

PR #932, opened 2024-09-10
🎨 display an error when the factory is infeasible

### LINKED ISSUE #918: ensure `/satisfy solve` is always feasible; or improve output when infeasible
Currently, an infeasible solution looks like:

image

We can ensure the solution is feasible by creating synthetic recipes to create all the raw resources (or every item, really) from nothing; ie a `Recipe("CreateIronOre", Building.Magic, input=NO_INPUT, output=IronOre * 1)`.  
But, this conflicts with `maximize` since we'd be able to infinitely create all items. That can be counteracted by giving those recipes an extremely negative objective value, so they're only used to maximize output if a given `output` cannot be reached.

The other option is just to leave it infeasible, and improve the output from `->` to something more usable.

twentylemon:
##### Summary

Fixes https://github.com/duck-dynasty/duckbot/issues/918

Went with yelling about the solution being infeasible. I may decide to revisit at a later date, /shrug.

##### Checklist

### DIFF
--- a/duckbot/cogs/games/satisfy/solver.py
+++ b/duckbot/cogs/games/satisfy/solver.py
@@ -1,9 +1,9 @@
 import itertools
 from functools import reduce
 from math import isclose
-from typing import Callable, List
+from typing import Callable, List, Optional
 
-from mip import INF, INTEGER, LinExpr, Model, Var, maximize, xsum
+from mip import INF, INTEGER, LinExpr, Model, OptimizationStatus, Var, maximize, xsum
 
 from .factory import Factory
 from .item import Item, sinkable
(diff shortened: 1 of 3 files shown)


duckbot/cogs/games/satisfy/solver.py
```diff
@@ -1,18 +1,24 @@
...
+from mip import INF, INTEGER, LinExpr, Model, OptimizationStatus, Var, maximize, xsum
 
 from .factory import Factory
 from .item import Item, sinkable
 from .recipe import ModifiedRecipe, Recipe
 
 zero = LinExpr(const=0)
 
+good_enough = [
+    OptimizationStatus.FEASIBLE,  # not optimal but /shrug
```
jameshughes89, line 15: 🚀

jameshughes89 (approved):

PR #935, opened 2024-09-11
🐛 improve rounding for output

twentylemon:
##### Summary

Moving the rounding to display instead of the solver return. Chopping off digits early makes the math wonky.

##### Checklist

### DIFF
--- a/duckbot/cogs/games/satisfy/pretty.py
+++ b/duckbot/cogs/games/satisfy/pretty.py
@@ -9,6 +9,7 @@ from .item import Item
 from .rates import Rates
 from .recipe import ModifiedRecipe
 
+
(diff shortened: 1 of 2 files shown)


duckbot/cogs/games/satisfy/pretty.py
```diff
@@ -10,8 +10,12 @@
 from .recipe import ModifiedRecipe
 
 
+def rnd(num: float) -> float:
```
jameshughes89, line 13: Why make your own though?

twentylemon: To keep the number of `4` digits in one place

jameshughes89 (approved):
Makes sense

PR #947, opened 2024-09-14
✨ make autocomplete check for letters in same order instead of substring

twentylemon:
##### Summary

Full substring is kinda annoying to search for. Changed so autocomplete searches for the same letters in the same order. So like `repl -> REinforcedironPLate`; capitals are the match. Example here as well:

{781878DE-58C5-4B81-96FA-6612F781BB1E}

##### Checklist


jameshughes89 (approved):
Neat use of an internal function

PR #961, opened 2024-09-22
🔧 weigh item usage by rough raw resource cost in objective function

### LINKED ISSUE #958: weigh item usage by rough raw resource cost in objective function
Current, objectives like maximize the output and minimize inputs used are aa basic linear combination based on the number of items per minute. This works fine if the inputs and outputs are roughly the same rates, but may break down for some of the more extremely dense or slow to produce items. 

Raw resources are weighed by their map availability. We could apply a similar weight to all input items based on their default recipe costs. We can't really use alternative recipe costs since the cardinality becomes huge for late game items, and we don't know how items which are inputs were produced. 

Weighing maximize by the highest resource usage for any item will also ensure that it is always the priority for a solution.

twentylemon:
##### Summary

Fixes https://github.com/duck-dynasty/duckbot/issues/958. The objective function is mostly unchanged by looking at it, but it should consider iron plates to be 1.5x more costly than iron ore, for example; instead of merely counting the rate of items produced. I wasn't able to think of a test case to show the theoretical bad behaviour. I may add some test cases in the future, or as we run into weird solves.

also fixed a bug in a recipe. #python.

##### Checklist


jameshughes89 (approved):
ok

PR #997, opened 2024-11-11
🐛 fix raw resource creation not working with intermediary item inputs

### LINKED ISSUE #995: 🐛  raw resource supply doesn't work with manual inputs
image

reason being [this objective](https://github.com/duck-dynasty/duckbot/blob/main/duckbot/cogs/games/satisfy/solver.py#L105)

twentylemon:
##### Summary

Fixes #995

Root cause ended up being a few layers. The symptom was that the solver would make a crapload of an input item even when it was not requested, since it could create new inputs and be rewarded for it (see link in issue). The original issue was that the weights for raw resource creation was wrong, so creating more of the input ended up blowing up the objective to infinity. My first stab at this was to bound the solution so I could inspect it. That's the `upper_bound` function introduced.   
After fixing the weights, it'd work so long as only default recipes were used (since weights are all based on default recipe resource costs). But, if cheaper alternates were included, the solver would prefer to use resources to create those since it was rewarded (cost N to make resources, but then would earn >N to make the item). I introduced a penalty to creating any items it did not have to. That's the `excess_target` objective change.  
Finally, after that, the solver seemed to really like creating power instead of leaving the factory simpler. I made the auxiliary items have max weight, instead of whatever weight they were calculated to be.

Note, a lot of this is arbitrary. But, tests pass, and we can always change weights, etc, later.

##### Checklist

### DIFF
--- a/duckbot/cogs/games/satisfy/rates.py
+++ b/duckbot/cogs/games/satisfy/rates.py
@@ -22,6 +22,11 @@ class Rates:
     def get(self, key: Item, default: Optional[float]) -> Optional[float]:
         return self.rates.get(key, default)
 
+    def singleton_rate(self) -> float:
+        if len(self.rates) != 1:
+            raise AssertionError(f"{self} is not a singleton")
+        return next(x for x in self.items())[1]
+
     def __bool__(self) -> bool:
         return bool(self.rates)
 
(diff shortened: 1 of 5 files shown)


jameshughes89:
> the solver seemed to really like creating power instead of leaving the factory simpler

Do you mean, it would suggest making, like, grid power? Why?

jameshughes89 (approved):

twentylemon:
> > the solver seemed to really like creating power instead of leaving the factory simpler
> 
> Do you mean, it would suggest making, like, grid power? Why?

It's based on item weights, I think. And that you can produce a lot of power from relatively little inputs. But it's one of those things where solver is as solver does.

PR #1003, opened 2024-11-13
✨ add function to calculate item weights by cheapest possible cost

twentylemon:
##### Summary

It's not used yet in the solver, I couldn't get it to work as I wanted. It also seems to be a little wrong still, since `StitchedIronPlate`, for example, should probably be cheaper than the default recipe for super plates. It may be an `approx` thing, but I haven't looked that deeply into it yet. The alternate weights also ended up producing some wonky solutions in only ways optimizers can; it'd find extremely convoluted solutions that use many different recipes which would end up minimizing the new costs (though, only with non-raw factory inputs, the typical input+maximize still works fine).

I added two solver tests to show the issue that I was trying to resolve here. It is an edge case, at any rate.

Note that the changes to the `solver` file are only moving the declarations to a new file. `alternate_weights` is the only real new code (some minor alterations otherwise, but only cleanup [moving auxiliary item weighing, `all_items` addition in `minmax_weights` function).

##### Checklist

### DIFF
--- a/duckbot/cogs/games/satisfy/item.py
+++ b/duckbot/cogs/games/satisfy/item.py
@@ -202,6 +202,9 @@ class Item(Enum):
 
         return Rates({self: rhs})
 
+    def __lt__(self, rhs):
+        return self.name < rhs.name
+
     def __str__(self):
         return self.name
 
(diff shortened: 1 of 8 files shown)


jameshughes89 (approved):
🚚

PR #1006, opened 2024-11-20
🐛 send multiple messages if solution is too big

### LINKED ISSUE #959: produce some kind of result when the solution factory is large
Discord embeds only support 25 fields, and currently 1 field is used per recipe. For larger factories, the command just crashes once it goes to send the message. The solver itself finds a solution just fine. 

Seeing large factories is still useful for planning, so maybe a basic text output for these is better than nothing.

twentylemon:
##### Summary

Fixes #959

Discord has a message size limit, which sometimes big solutions will run into. This chunks up the message into the biggest possible components that can fit into a single message, then send multiple messages.

##### Checklist

### DIFF
--- a/duckbot/util/embeds/__init__.py
+++ b/duckbot/util/embeds/__init__.py
@@ -1,3 +1,7 @@
+from typing import List
+
+from discord import Embed
+
 # https://discord.com/developers/docs/resources/channel#embed-limits
 MAX_EMBED_LENGTH = 6000
 MAX_TITLE_LENGTH = 256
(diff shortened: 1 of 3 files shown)


duckbot/util/embeds/__init__.py
```diff
@@ -7,3 +11,18 @@
...
+    total, group, groups = 0, [], []
+    for embed in embeds:
+        length = len(embed)
+        if total + length < MAX_EMBED_LENGTH:
+            total += length
+            group.append(embed)
+        else:
+            total = 0
+            groups.append(group)
+    return groups + [group] if group else groups
```
jameshughes89, line 28: jesus

tests/util/embeds/group_by_max_length_test.py
```diff
@@ -0,0 +1,34 @@
...
+from discord import Embed
+
+from duckbot.util.embeds import group_by_max_length
+
+
+def embed(length: int) -> Embed:
+    return Embed(title="a" * length)
+
+
+@pytest.mark.parametrize("length", [0, 1, 3000, 6001])
```
jameshughes89, line 11: are 3000 and 6001 meaningful?

jameshughes89: https://github.com/duck-dynasty/duckbot/pull/1006/files#diff-852d46fbcd57025c42fb5af4c1d67bca90bcaf8ad58ab1362b17cbda69194db3L2

OK, I see

twentylemon: 6000 is the max embed length discord has. 3000 is just a non-1 number within the size limit

jameshughes89 (approved):

duckbot/util/embeds/__init__.py
twentylemon: I imagine there's a functional way to do this, but I can't think of it right now.

PR #1008, opened 2024-11-20
🐛 fix state error in embed grouping calculation

twentylemon:
##### Summary

Related to #1006. Logic there missed reseting the `group` variable. Dang ol' state.

Refactored the tests a bit when adding the new one. Changed the `embed` creation function to include the size so errors were easier to read, but it means there is no 0 length embed any more. /shrug.

##### Checklist

### DIFF
--- a/duckbot/util/embeds/__init__.py
+++ b/duckbot/util/embeds/__init__.py
@@ -19,7 +19,7 @@ def group_by_max_length(embeds: List[Embed]) -> List[List[Embed]]:
     total, group, groups = 0, [], []
     for embed in embeds:
         length = len(embed)
-        if total + length < MAX_EMBED_LENGTH:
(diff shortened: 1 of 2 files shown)


duckbot/util/embeds/__init__.py
```diff
@@ -23,6 +23,6 @@ def group_by_max_length(embeds: List[Embed]) -> List[List[Embed]]:
             total += length
             group.append(embed)
         else:
-            total = 0
             groups.append(group)
+            total, group = 0, [embed]
```
jameshughes89, line 27: Why one line though?

jameshughes89 (approved):

duckbot/util/embeds/__init__.py
twentylemon: why not? it groups "reset state" in one line

PR #1066, opened 2025-02-01
:fire: remove release action

twentylemon:
##### Summary

After having this for quite some time, it doesn't seem to provide any tangible benefit. Originally I thought we'd only need to say `release` is a required check for the branch, but all actions show up in the UI anyways. That and we need to update dependencies of `release` when a new job was added, which is the same as updating `deploy` with this change. Removing also means we don't run some code in the cloud, so sure, save power or whatever.

I've updated the needed checks on the repo already to be the 3 actions here.

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -59,21 +59,11 @@ jobs:
         . ../${{ steps.cdk.outputs.venv-path }}/bin/activate
         cdk synth
 
-  release:
+  deploy:
     needs:
     - test
     - sanity
     - synthesize
-    runs-on: ubuntu-latest
-    if: ${{ always() }}
-    steps:
-    - run: exit 1
-      if: ${{ contains(needs.*.result, 'failure') }}
-    - run: echo all checks passed
-
-  deploy:
-    needs:
-    - release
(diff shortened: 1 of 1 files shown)


.github/workflows/python-package.yml
```diff
@@ -59,22 +59,12 @@ jobs:
...
-    - run: exit 1
-      if: ${{ contains(needs.*.result, 'failure') }}
-    - run: echo all checks passed
-
-  deploy:
-    needs:
-    - release
-    runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.repository == 'duck-dynasty/duckbot'
     concurrency: deploy
```
jameshughes89, line 69: Is this needed?

twentylemon: which part? github trying to be helpful with the context diff and it's not working. `concurrency: deploy`? if so, that makes it so only one instance of the job can run at once, and yes, we only want one deployment running at a time

jameshughes89: It was the concurrency deploy (blame mobile)

jameshughes89 (approved):

PR #1067, opened 2025-02-01
🚀 add prod env to deploy job

twentylemon:
##### Summary

For some reason, doing lots of deployments back to back causes the ec2 instance to fall out of sync with the ecs service that is using the instance as a deployment target. The instance is there and running, but ecs loses track of it and just stalls forever. I had to manually replace the ec2 instance twice now, then ecs is fine. It seems to happen when using the merge queue for these dependabot updates.

I created the prod GitHub environment and slapped on a 15min wait timer. That means this actions job will wait for 15 minutes before starting, which hopefully should give everything in aws to settle before deploying again.

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -77,6 +77,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.repository == 'duck-dynasty/duckbot'
     concurrency: deploy
+    environment: prod
     steps:
     - uses: actions/checkout@v4
     - name: Build and Push Docker Image


jameshughes89:
Where does the prod GitHub environment exist? I see the `environment: prod` in the action, but where is prod from?

twentylemon:
Are you a repo admin? It's settings -> environemnts

jameshughes89 (approved):

jameshughes89:
> Are you a repo admin? It's settings -> environemnts

I don't believe I am. Either way, sounds like.it does in fact exist somewhere.

PR #1082, opened 2025-03-23
✨ randomize insights timer on boot

twentylemon:
##### Summary

With duckbot rebooting daily, the timer for replying to insightful messages is more predictable. This changes the loop to pick a random interval instead of only 139 minutes. It's set only on startup once, but it's still less predictable this way.

##### Checklist


jameshughes89 (approved):
Those are some prime numbers you got there

PR #1103, opened 2025-05-13
:alien: address backward incompatible `setuptools` changes

twentylemon:
##### Summary

Fixing issues like this one: https://github.com/duck-dynasty/duckbot/actions/runs/14770581040/job/41469833019?pr=1100#step:4:535

Two things here,
* `develop` is no longer a valid command, it's been replaced by `editable_wheel`
  * we rely on this to install nltk data. `install` still works, but we want editable installs for local work
  * though, this means we don't need the legacy param crap any more, so that's nice
* scripts apparently need to be in a python module now, importing directly from `setup` doesn't work

##### Checklist


jameshughes89 (approved):
__init__

PR #1201, opened 2025-11-06
:bug: pyowm --- Update to 3.4 and fix imports

jameshughes89:
##### Summary

Related to #1152

1. Update to use pyowm 3.4
2. Update the imports that were `weatherapi25` -> `weatherapi30` (2.5 does not work anymore with update to >=3.4)
3. Fixed some type hints
4. Updated the one_call method to work with the update to pyowm 

##### Checklist

I ran locally with docker and was able to get it working with version 3.4. Version 3.5 seems to have caused some breaking changes with the city lookup thing. I did not fully look into this 3.5 issue yet and it seems like there could be an issue in a submodule, but whatever, here is a working version with 3.4. 

I think going from 3.3 -> 3.4 is an improvement as the changes here will make it work with versions >=3.4. If a version 3.5.1 comes out where the city lookup works fine, then great. Until then, we are stuck on 3.4.

### DIFF
--- a/duckbot/cogs/weather/weather.py
+++ b/duckbot/cogs/weather/weather.py
@@ -36,9 +36,7 @@ class Weather(commands.Cog):
         return self._owm
 
     def one_call(self, **kwargs):
-        # pyowm doesn't support onecall 3.0: https://github.com/csparpa/pyowm/issues/404
-        json = requests.get(url="https://api.openweathermap.org/data/3.0/onecall", params=kwargs | {"appid": os.getenv("OPENWEATHER_TOKEN")}).json()
-        return OneCall.from_dict(json)
+        return self.weather_manager().one_call(**kwargs)
 
     @commands.hybrid_group(name="weather", invoke_without_command=True)
     async def weather_command(self, context: commands.Context, city: Optional[str] = None, country: Optional[str] = None, index: Optional[int] = None):
--- a/duckbot/cogs/weather/weather.py
+++ b/duckbot/cogs/weather/weather.py
@@ -12,8 +12,8 @@ import requests
 import timezonefinder
 from discord.ext import commands
 from pyowm.utils import config
-from pyowm.weatherapi25.location import Location
-from pyowm.weatherapi25.one_call import OneCall
+from pyowm.weatherapi30.location import Location
+from pyowm.weatherapi30.one_call import OneCall
 
 from duckbot.db import Database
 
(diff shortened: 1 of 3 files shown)


twentylemon (commented):
I was fiddling with this a while ago and ran into some issues with the location/city search. Does that just work now? Have you tried it out?

jameshughes89:
> I was fiddling with this a while ago and ran into some issues with the location/city search. Does that just work now? Have you tried it out?

I did not test that. I do not have an open weather token. 

I can get one, but wont be able to tackle this for a bit. If curious, give it a go locally.

jameshughes89:
> > I was fiddling with this a while ago and ran into some issues with the location/city search. Does that just work now? Have you tried it out?
> 
> I did not test that. I do not have an open weather token.
> 
> I can get one, but wont be able to tackle this for a bit. If curious, give it a go locally.

I got a token and have it running. Setting locations is borked as you said.

jameshughes89:
> > > I was fiddling with this a while ago and ran into some issues with the location/city search. Does that just work now? Have you tried it out?
> > 
> > 
> > I did not test that. I do not have an open weather token.
> > I can get one, but wont be able to tackle this for a bit. If curious, give it a go locally.
> 
> I got a token and have it running. Setting locations is borked as you said.

Something funny happened with version 3.5 where the locations become a problem. I tested locally with version 3.4 and it seems to work fine. 

I will revert to 3.4 (still an upgrade from 3.3) as it still makes the changes that will be needed going forward.

twentylemon:
> I think going from 3.3 -> 3.4 is an improvement

So long as we're using onecall v3, seems like we'd be safe from openweather shutting down v2.5. Definitely an improvement.

twentylemon (dismissed):

jameshughes89:
> > I think going from 3.3 -> 3.4 is an improvement
> 
> So long as we're using onecall v3, seems like we'd be safe from openweather shutting down v2.5. Definitely an improvement.

I want to do one more test locally before I merge. I tested that I can set a location, but I never tried to plot the graph. I have no reason to believe it would not work, but nevertheless, I want to check.

jameshughes89:
Something is now not working with `get`. Set PR to Draft. 
image

jameshughes89:
I will try adding a `self.weather_manager = self.omw.weather_manager` to the `Weather` class. That may be the trick.

jameshughes89:
The trouble now is that using One Call version 3.0 requires a subscription that, although free for 1000 calls, still requires CC details, which is no good. Version 2.5 which we were using does not require the CC. 

I am currently looking into alternatives.

jameshughes89:
Dead branch; at best there could be a new branch with an alternative to `pyowm`

PR #1228, opened 2026-01-06
🚀 update ec2 ami to address AL2 end of life

twentylemon:
##### Summary

**NOTE** I'll merge in the morning. If we deploy when the host is shut down overnight, it just gets stuck and requires some manual intervention.

AL2 is going end of life in June. This new ami uses the recommended AL2023, which says it's supported until 2028 some time. Should be no affect otherwise; we did this same when I was still working at Amazon and it always sent smooth there. The only thing for us is that an 8gb drive might not be enough any more, we'll see.

##### Checklist


jameshughes89 (approved):

jameshughes89:
> NOTE I'll merge in the morning. If we deploy when the host is shut down overnight, it just gets stuck and requires some manual intervention.

:+1:

PR #1283, opened 2026-03-07
🗃️ set up alembic and drop `city_id` column

### LINKED ISSUE #1282: set up alembic to drop the `city_id` column from the saved weather locations table
related https://github.com/duck-dynasty/duckbot/pull/1281 https://github.com/duck-dynasty/duckbot/pull/1152

pyowm upgrade removed cities from their API, so we are using geolocations now which only have lat/lon. We have a non-nullable `city_id` in our db though, so new `/weather set`s are failing on write.

example logs
> DETAIL: Failing row contains (id, city, CA, null, lat, lon).
> psycopg2.errors.NotNullViolation: null value in column "city_id" of relation "weather_locations" violates not-null constraint

twentylemon:
##### Summary

fixes https://github.com/duck-dynasty/duckbot/issues/1282

tested locally, results in [dev channel](https://discord.com/channels/845352633702809621/859095806878941216/1479892517032820900)

##### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ ENV VIRTUAL_ENV "/opt/venv"
 ENV PATH "$VIRTUAL_ENV/bin:$PATH"
 WORKDIR /duckbot
 COPY --from=pip-dependencies "$VIRTUAL_ENV" "$VIRTUAL_ENV"
+COPY alembic.ini .
 COPY resources/ ./resources
 COPY duckbot/ ./duckbot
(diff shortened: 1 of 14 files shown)


twentylemon:
sanity failing with
> sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) could not translate host name "postgres" to address: Temporary failure in name resolution

No db locally there. I'll if we can use the whole compose stack here, makes sense to sanity check the migrations too.

Chippers255 (approved):

PR #1310, opened 2026-05-01
:bug: have `now` incorporate `year` for duckbot inception day

### LINKED ISSUE #642: Have `now` incorporate `year`
From #640 and #641. 

It was observed by @twentylemon that the calculation for the time duckbot was alive ([found here](https://github.com/duck-dynasty/duckbot/blob/2cae4ecc4402812cb9067ae6e2f10d538bb250c0/duckbot/cogs/announce_day/special_days.py#L52)) has a potential problem since `year` is a parameter for the special days list but the `now` calculation ([found here](https://github.com/duck-dynasty/duckbot/blob/2cae4ecc4402812cb9067ae6e2f10d538bb250c0/duckbot/cogs/announce_day/datetime.py#L14-L19)) does not make use of the `year`. 

As additionally noted by @twentylemon:

> So what I mean here for year, this method mutates this class, adding dates to some internal map like {"2020-01-01": ["New Year's"]}. This method is lazily called when you ask for a date in a particular year.
> So let's imagine a typical use case.
> ```python
> days = holidays.get(today_in_2022)
> # ... the next day
> days = holidays.get(tomorrow_in_2022)
> ```
> The first call ends up populating the special days for 2022. DuckBot's inception day gets some value like "I'm n seconds old". In the second call, there's no need to populate the days again, so it is not called. Say tomorrow is DuckBot's special day, we end up saying "I'm n seconds old", but it's actually n+1 day seconds. now() is not recalculated.
> DuckBot currently turns off every day, but already was not the case in the past, so we can't rely on _populate being called daily. We also can't force call it, since self[date] = x actually appends to the internal list of days the class keeps. now() needs to somehow be based on the year. We could just say datetime(year, 12, 3, 7, 0), but that also loses accuracy. Not sure what the best approach is here.

jameshughes89:
##### Summary

Fixes #642. `holidays._populate(year)` runs once per year and caches its result, so the Dec 3 inception-day string was baking a stale `now()` snapshot into the map. Moved the duration formatting into a `get_list` override so the seconds-since-inception is computed at lookup time, not populate time.

##### Checklist

### DIFF
--- a/duckbot/cogs/announce_day/special_days.py
+++ b/duckbot/cogs/announce_day/special_days.py
@@ -6,10 +6,7 @@ from dateutil.relativedelta import FR, MO, SU
 from dateutil.relativedelta import relativedelta as rd
 from holidays.countries import CA
 
-from duckbot.util.datetime import now, timezone
-
-INCEPTION = datetime(2020, 12, 3, 10, 39, tzinfo=timezone())
-DUCKBOT_DAY_TEMPLATE = "DuckBot's Inception Day. I'm about {seconds}s old"
+from duckbot.util.datetime import timezone
 
 
 class SpecialDays(CA):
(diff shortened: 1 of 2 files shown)


twentylemon:
Thinking we should just drop this and close the issue. Duckbot isn't going to be on 24/7 any time soon.

jameshughes89:
> Thinking we should just drop this and close the issue. Duckbot isn't going to be on 24/7 any time soon.

¯\\_(ツ)_/¯

PR #1327, opened 2026-05-01
fix: convert underscore italics to asterisk italics in haiku embed

### LINKED ISSUE #1311: underscore-italic markdown in a haiku breaks the embed's wrapping italic
Since #1309 the haiku embed preserves the user's markdown, which works for `**bold**`, `*asterisk italic*`, `~~strike~~`, `` `code` ``, and `||spoiler||`. It does **not** work for `_underscore italic_`, because [`detect_haiku`](https://github.com/duck-dynasty/duckbot/blob/main/duckbot/cogs/messages/haiku.py#L31) wraps the whole haiku in `_..._` to italicize it inside the embed, and the user's underscores collide with the wrapper.

Example: a message like `_five_ seven five` produces an embed value of `__five_\nseven\nfive_`, where the leading `__` and the inner `_` get reparsed by Discord and the formatting comes out wrong.

A few directions, no preference:
- Wrap the haiku in something other than `_..._` (e.g. asterisks, or use the embed's italic styling).
- Escape underscores in the haiku text before wrapping.
- Strip just the underscore-italic flavor while preserving other markdown.

jameshughes89:
## Summary

- `detect_haiku` wraps the haiku in `_..._` to italicise it in the embed, but user messages containing `_underscore italic_` cause the underscores to collide with the wrapper (e.g. `_five_ seven five` → embed value `__five_\nseven\nfive_`, where `__` is parsed as underline by Discord).
- Fix: convert `_word_` italics to `*word*` italics in the assembled haiku before building the embed value. Both render as italic in Discord, but `*` doesn't collide with the wrapping `_..._`. Double-underscore (`__word__`) is left alone via lookbehind/lookahead so underline is preserved.
- Added test cases for underscore-italic input (expecting asterisk-italic output) and other markdown variants.

Closes #1311.

## Test plan

- [x] `pytest tests/cogs/messages/haiku_test.py` — 65 tests, all pass
- [x] `flake8` clean on changed files
- [x] `black` / `isort` — no reformatting needed

### DIFF
--- a/duckbot/cogs/messages/haiku.py
+++ b/duckbot/cogs/messages/haiku.py
@@ -1,7 +1,11 @@
+import re
+
 from discord import Colour, Embed
 from discord.ext import commands
 from discord.utils import remove_markdown
 from nltk.corpus import cmudict
+
+UNDERSCORE_ITALIC_RE = re.compile(r"(?<!_)_([^_\n]+?)_(?!_)")
 
 
 class Haiku(commands.Cog):
@@ -27,7 +31,8 @@ class Haiku(commands.Cog):
             else:
                 return  # it's not a haiku
         if i == len(words):
-            haiku = "\n".join(lines).replace("_", r"\_")
+            # Convert `_..._` italics to `*..*` so user underscores don't collide with the wrapping italic.
+            haiku = UNDERSCORE_ITALIC_RE.sub(r"*\1*", "\n".join(lines))
             embed = Embed(colour=Colour.dark_red()).add_field(name=":cherry_blossom: **Haiku Detected** :cherry_blossom:", value=f"_{haiku}_")
             await message.channel.send(embed=embed)
 
--- a/duckbot/cogs/messages/haiku.py
+++ b/duckbot/cogs/messages/haiku.py
@@ -5,7 +5,7 @@ from discord.ext import commands
 from discord.utils import remove_markdown
 from nltk.corpus import cmudict
 
-UNDERSCORE_ITALIC_RE = re.compile(r"(?<!_)_([^_\n]+?)_(?!_)")
+UNDERSCORE_ITALIC_RE = re.compile(r"(?<!\w)_([^_\n]+?)_(?!\w)")
 
 
 class Haiku(commands.Cog):
--- a/duckbot/cogs/messages/haiku.py
+++ b/duckbot/cogs/messages/haiku.py
@@ -27,7 +27,7 @@ class Haiku(commands.Cog):
             else:
                 return  # it's not a haiku
         if i == len(words):
-            haiku = "\n".join(lines)
+            haiku = "\n".join(lines).replace("_", r"\_")
             embed = Embed(colour=Colour.dark_red()).add_field(name=":cherry_blossom: **Haiku Detected** :cherry_blossom:", value=f"_{haiku}_")
             await message.channel.send(embed=embed)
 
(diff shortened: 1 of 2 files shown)


duckbot/cogs/messages/haiku.py
```diff
@@ -27,7 +27,7 @@ async def detect_haiku(self, message):
             else:
                 return  # it's not a haiku
         if i == len(words):
-            haiku = "\n".join(lines)
+            haiku = "\n".join(lines).replace("_", r"\_")
```
twentylemon: This just keeps the underscores. We want italics

jameshughes89:
Swapped the escape for a regex that rewrites `_word_` → `*word*` in the assembled haiku. Both render as italic in Discord, so the formatting is preserved, but `*` doesn't collide with the wrapping `_..._`. Double-underscore (underline) is left alone via lookbehind/lookahead, so `__word__` still renders as underline.

duckbot/cogs/messages/haiku.py
jameshughes89: (Really me)

Currently we wrap the whole haiku in italics (`value=f"_{haiku}_"`). 

So, this would require either (a) drop wrapping the whole thing in italics so we can have internal parts being *italics* or (b) just ignore italics formatting for anything internal.

tests/cogs/messages/haiku_test.py
```diff
@@ -54,11 +54,13 @@ async def test_detect_haiku_finds_case_insensitive_haiku(message):
...
         ("**five** seven five", "**five**\nseven\nfive"),
-        ("*five* _seven_ five", "*five*\n_seven_\nfive"),
+        ("*five* seven five", "*five*\nseven\nfive"),
+        ("_five_ seven five", "*five*\nseven\nfive"),
+        ("*five* _seven_ five", "*five*\n*seven*\nfive"),
         ("~~five~~ seven five", "~~five~~\nseven\nfive"),
         ("`five` seven five", "`five`\nseven\nfive"),
         ("||five|| seven five", "||five||\nseven\nfive"),
-        ("**five**, _seven_! ~~five~~.", "**five**\n_seven_\n~~five~~"),
+        ("**five**, _seven_! ~~five~~.", "**five**\n*seven*\n~~five~~"),
```
twentylemon, line 63: Seems to remove the `!` as well? Is that removed by `remove_markdown` ?

duckbot/cogs/messages/haiku.py
```diff
@@ -1,8 +1,12 @@
+import re
+
 from discord import Colour, Embed
 from discord.ext import commands
 from discord.utils import remove_markdown
 from nltk.corpus import cmudict
 
+UNDERSCORE_ITALIC_RE = re.compile(r"(?<!_)_([^_\n]+?)_(?!_)")
```
twentylemon: I don't read regular expression, does this account for non-italics like your_mother?

tests/cogs/messages/haiku_test.py
jameshughes89: Pre-existing — `get_words` strips `!` (and `,`, `.`, `?`) before syllable counting. Not related to this change.

duckbot/cogs/messages/haiku.py
jameshughes89: Fixed — switched to `(?<!\w)` and `(?!\w)` so underscores only match at word boundaries, same as how Discord itself parses italic markdown. `snake_case_var` no longer matches.

duckbot/cogs/messages/haiku.py
jameshughes89: Went with a hybrid: keep the outer `_..._` wrapper, but rewrite internal `_word_` → `*word*` so internal italics still render. Avoids changing the overall haiku styling while still preserving user formatting. Happy to switch to (a) — drop the wrapper — if you prefer; would simplify the code (no regex needed).

twentylemon (approved):

PR #1331, opened 2026-05-14
feat: add bold and italic to mock text output

### LINKED ISSUE #247: add italics/bold to text generated by `!mock`

jameshughes89:
Adds per-character bold/italic formatting to `!mock`, chosen randomly per character.

Discord's markdown parser collapses adjacent asterisk spans (e.g. `**H****i**` renders as `***H***i**` — broken), so emphasis spans are separated by at least one plain character to prevent collision.

Closes #247

---

Will test locally to confirm Discord renders correctly before merging.

### DIFF
--- a/duckbot/cogs/text/mock_text.py
+++ b/duckbot/cogs/text/mock_text.py
@@ -1,11 +1,10 @@
 import random
-import re
 
 from discord.ext import commands
 
 import duckbot.util.messages
 
-WRAPPERS = ["", "**", "_"]
+WRAPPERS = ["", "**", "*"]
 
 
 class MockText(commands.Cog):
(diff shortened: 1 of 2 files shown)


twentylemon:
> * wraps each word-bearing token with `**bold**`, `_italic_`, or nothing (uniform 1/3 each)

I envisioned this more as mixed character by character, not word by word. It's more typical that we mock a single word anyways

jameshughes89:
Good call — switched to per-character in 90e0de8. Each alpha char independently picks bold / italic / plain. Also switched italic from `_` to `*` so it actually renders mid-word (`_x_` only italicizes at word boundaries in Discord; `*x*` works anywhere).

jameshughes89:
image

Looks good.

twentylemon:
> image
> Looks good.

I wonder, would it look better to have the formatters be longer than one character each? Algorithm similar to this might make sense:
1. split the string into a random number of non-empty spans
2. assign each span a format (or any combination of formats), removing the prior format as a choice for this one

jameshughes89:
> > image
> > Looks good.
> 
> I wonder, would it look better to have the formatters be longer than one character each? Algorithm similar to this might make sense:
> 
> 1. split the string into a random number of non-empty spans
> 2. assign each span a format (or any combination of formats), removing the prior format as a choice for this one

image

I added `backticks` to mock too for more variety. 

It gives me magazine cut out letters ransom note vibes.

jameshughes89:
i hit close with comment by accident

twentylemon (approved):

PR #1332, opened 2026-05-14
feat: crit hit/fail flavour on single d20 rolls

### LINKED ISSUE #196: add critical hit/fail flavour to `!roll`
See original issue #128, rolling dice was added in #188 but no flavour was added.

jameshughes89:
Adds critical hit/fail flavour text to `!roll` when a single `1d20` rolls a natural 20 or 1.

Does not conflict with the natural 20 advantage/disadvantage mechanic — that feature would apply before the roll, while this only reads the final rolled value.

Closes #196

---

Will test locally to confirm the flavour text appears correctly in Discord before merging.

### DIFF
--- a/tests/cogs/games/dice_test.py
+++ b/tests/cogs/games/dice_test.py
@@ -1,8 +1,10 @@
 from unittest import mock
 
 import d20
+import pytest
 
 from duckbot.cogs.games import Dice
+from duckbot.cogs.games.dice import CRIT_FAIL_FLAVOUR, CRIT_HIT_FLAVOUR
 from duckbot.util.messages import MAX_MESSAGE_LENGTH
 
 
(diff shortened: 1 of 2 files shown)


jameshughes89:
Open question: should the crit flavour stay d20-only, or generalize to any single-die roll hitting its natural min/max?

- **d20-only (current behaviour)**: matches D&D convention; `1d6` rolling a 6 doesn't say "critical".
- **Generalized**: `1d6` → 6 or 1 also gets flavour; same for `1d100`, etc. Would need different wording (probably "Maximum roll!" / "Minimum roll!" instead of D&D's "critical").

Happy to follow up with the generalized version if you want it.

tests/cogs/games/dice_test.py
```diff
@@ -35,3 +35,51 @@ async def test_roll_sends_result(roller, context):
...
     await clazz.roll(context, "1d20")
     context.send.assert_called_once_with("**Rolls**: results\n**Total**: 1")
+
+
+@mock.patch("random.randrange", return_value=19)  # randrange(20) + 1 == 20
+async def test_roll_natural_20_shows_crit_hit(randrange, context):
+    clazz = Dice()
+    await clazz.roll(context, "1d20")
+    sent = context.send.call_args[0][0]
+    assert ":dart: **Critical hit!**" in sent
```
twentylemon: match the testing style -- assert the mock `context.send.assert_called_once_with` ; prefer parameterized tests where possible to reduce the noise

jameshughes89:
image

image

twentylemon (approved):

PR #1336, opened 2026-05-18
:sparkles: toggle `/satisfy maximize` on repeat invocations

twentylemon:
##### Summary

Re-running `/satisfy maximize` on an item that is already in the factory's maximize set now removes it, instead of being a silent no-op (set semantics). This removes the need to `/satisfy reset` the entire factory just to drop a single maximize item. Renamed the handler from `add_maximize` to `toggle_maximize` to reflect the new behaviour; the slash command name is unchanged. Wiki docs updated to reflect the new behaviour (previously claimed a full reset was required).

Tested manually by running the new unit test; relied on CI for the full suite since my local env isn't fully set up.

##### Checklist

### DIFF
--- a/duckbot/cogs/games/satisfy/satisfy.py
+++ b/duckbot/cogs/games/satisfy/satisfy.py
@@ -66,7 +66,7 @@ class Satisfy(Cog):
         self.save(context, factory)
         await context.send(embed=factory_embed(factory), delete_after=60)
 
-    @satisfy.command(name="maximize", description="Toggle whether to maximize output of the given item.")
+    @satisfy.command(name="maximize", description="Toggle maximizing output of the given item.")
     async def toggle_maximize(self, context: Context, item: str):
         factory = self.factory(context)
         if Item[item] in factory.maximize:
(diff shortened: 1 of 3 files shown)


duckbot/cogs/games/satisfy/satisfy.py
```diff
@@ -66,10 +66,13 @@ async def add_target(self, context: Context, item: str, rate_per_minute: float):
...
-    @satisfy.command(name="maximize", description="Specify maximize output of desired item.")
-    async def add_maximize(self, context: Context, item: str):
+    @satisfy.command(name="maximize", description="Toggle maximizing output of the given item.")
+    async def toggle_maximize(self, context: Context, item: str):
         factory = self.factory(context)
-        factory.maximize.add(Item[item])
+        if Item[item] in factory.maximize:
+            factory.maximize.discard(Item[item])
+        else:
+            factory.maximize.add(Item[item])
```
jameshughes89, line 75: You may be able to use something like `^=` here (symmetric difference). 

HOWEVER, to be clear, I prefer this (this being the if/else).

twentylemon: I'm usually pro set ops but I prefer the function calls for mutations instead of re-assignment

jameshughes89 (approved):
Potential test --- `test_toggle_maximize_removes_toggled_only` to check a case where a set has many items, one is toggled, and everything else in set is left alone.

jameshughes89 (approved):

PR #1433, opened 2026-07-21
:sparkles: add !meme command reposting a random attachment from #toms-memes

### LINKED ISSUE #57: Add `!meme` to display a random meme from a meme repository
I'm sure Tom has a truckload somewhere. We might be able to give the bot access to google photos to fetch a random image. Or, perhaps just pull something from the top of reddit or something.

twentylemon:
##### Summary

Adds `!meme` / `/meme`, which reposts a random image from #toms-memes without citing the author. _fixes #57_

Discord gives no way to sample channel history uniformly, so this uses the `around` approach from the issue thread: pick a uniformly random instant between the channel's creation and now, pull 100 messages around it, keep the ones with attachments, and post one attachment's URL. A message with several attachments contributes a random one. A window can land in a dead stretch with no attachments at all, so up to 5 windows are tried before falling back to a gnocchi soup gif — which also covers the channel not being found.

The cog lives in `cogs/insights` rather than its own package: both cogs post something amusing into the Friends Chat sourced from channel history, and share the `utils.get` channel lookup. `phrases.py` is renamed to `insight_phrases.py` so the module says whose phrases it holds, matching `cogs/messages/touch_grass_phrases.py`.

Tested with unit tests only; the `around` sampling hasn't been exercised against the live channel yet.

##### Checklist


duckbot/cogs/insights/memes.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Memes(commands.Cog):
+    def __init__(self, bot: Client):
+        self.bot = bot
+
+    @commands.hybrid_command(name="meme", description="Show a random meme from the memes channel")
+    async def meme(self, context: commands.Context):
+        channel = self.get_memes_channel()
+        url = await self.find_meme(channel) if channel else None
+        await context.send(url if url else "https://tenor.com/view/gnocchi-soup-gif-27425983")
```
jameshughes89: Discord will timeout after 3s of nothing and will not ever send. Could maybe wrap it with `context.typing` so it doesnt timeout. 

```python
async with context.typing():
          channel = self.get_memes_channel()
          url = await self.find_meme(channel) if channel else None
          await context.send(url if url else
  "https://tenor.com/view/gnocchi-soup-gif-27425983")
```

twentylemon: Should probably cram this into the claude.md too, not the first time it was forgotten

jameshughes89 (approved):

PR #1438, opened 2026-07-23
:whale: install apt packages with --no-install-recommends, verified by CI smoke tests

twentylemon:
##### Summary

Install apt packages with `--no-install-recommends` in both stages of the `Dockerfile`. This skips recommended-but-not-required packages, shrinking the prod image by ~204MB (2.11GB → 1.91GB) and resolving the SonarQube `docker:S6500` findings.

To make sure a recommended package the bot actually depends on isn't silently dropped, the `sanity` CI job now builds the image and runs a smoke test per OS/runtime dependency the bot shells out to or renders with. Each check is its own step (with `if: ${{ !cancelled() && steps.build.outcome == 'success' }}` so they all run independently and a failure names the culprit):

- `fortune`, `cowsay`, `ffmpeg` — binaries, plus the exact `!fortune` pipeline
- `graphviz` — renders the real `/satisfy` solution graph through `dot`
- `matplotlib` — renders a plot to PNG (Agg backend)
- `mip` — builds and solves a model, exercising the bundled CBC solver
- `import duckbot`

**Testing:** Built both variants locally and diffed sizes. Verified every check passes on the healthy image and exits nonzero when its dependency is broken (missing binary → 127, broken cowsay data dir → 2, missing `dot`/Python package → 1).

##### Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -39,9 +39,37 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v7
+    - name: Build Image
+      id: build
+      run: docker compose build duckbot
+    - name: Verify fortune
+      if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      run: docker run --rm --entrypoint fortune duckdynasty/duckbot:latest -a
+    - name: Verify cowsay
+      if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      run: docker run --rm --entrypoint cowsay duckdynasty/duckbot:latest moo
+    - name: Verify ffmpeg
+      if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      run: docker run --rm --entrypoint ffmpeg duckdynasty/duckbot:latest -version
+    - name: Verify fortune command
+      if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      run: docker run --rm --entrypoint sh duckdynasty/duckbot:latest -ec 'fortune -a | cowsay -f "$(ls /usr/share/cowsay/cows/ | shuf -n1)"'
+    - name: Verify duckbot imports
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -43,9 +43,37 @@ jobs:
...
+      run: docker compose build duckbot
+    - name: Verify fortune
+      if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      run: docker run --rm --entrypoint fortune duckdynasty/duckbot:latest -a
+    - name: Verify cowsay
+      if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      run: docker run --rm --entrypoint cowsay duckdynasty/duckbot:latest moo
+    - name: Verify ffmpeg
+      if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      run: docker run --rm --entrypoint ffmpeg duckdynasty/duckbot:latest -version
```
jameshughes89, line 57: Only checks that it loads. Doesn't check decoding MP3, vol filter, or emitting to stdout. Comparing to something else like the graphviz, it does a deeper check. Could do something like this. 

```
    run: >
      docker run --rm --entrypoint ffmpeg duckdynasty/duckbot:latest
      -v error -i resources/who-can-it-be-now.mp3
      -filter:a "volume=0.125" -f s16le -ac 2 -ar 48000 -
      | head -c 1 | wc -c | grep -q 1
```

twentylemon: I don't see why this would be necessary. If ffmpeg is there, it's there.

jameshughes89: Could the graphviz one be similarly changed to simplify it? check that it exists?

twentylemon: Graphviz by comparison is using the python wrapper, which require more setup to get to actually executing the binary underneath

jameshughes89 (approved):

PR #1484, opened 2026-08-28
:sparkles: Carry open markets across the season rollover

twentylemon:
##### Summary

Season rollover currently force-voids every still-`OPEN` market, refunds the stakes and deletes the positions. Any question with a horizon longer than the quarter it was asked in dies on the calendar rather than because someone answered it, and so does every market whose creator went quiet.

Open markets and their positions now ride into the next season instead. The obstacle is that a bet is paid with Season N coins and would pay out in Season N+1 coins, so the carry is charged at **cost**: rollover stops *refunding* your stake and debits it from the new season's grant instead. Bet 1,000 on a market that survives the reset and you open the next season with 9,000 coins plus the position, regardless of what the odds have done since — you are never charged for a gain you have not cashed in. Selling frees the coins back up.

The carried amount is exactly what the forced void used to refund (`max(0, -_invested(...))`), which keeps the rest of the mechanics honest for free: a later void of a carried market refunds precisely what was debited, so `_invested` needs no season scoping and no signature change, and a market carried across two boundaries is charged again each season and still voids square.

Two smaller changes fall out of it:

- `SeasonResult.final_balance` switches from cash to net worth (cash + mark-to-market share value). Under a cost-basis carry the old cash-only snapshot would record a player who put 9,000 into a market that carries forward as having finished the season with 1,000 coins. Ranking on net worth also makes the archived standings agree with the `/market leaderboard` people were watching seconds earlier, which already ranks that way.
- `void_resolved`'s archived-season guard now permits admin corrections on a carried market, since it belongs to the live season.

Worth noting for review:

- No schema change and no migration — the `season_id` reassignment is a data update and `reason` is free-text. The new `carry` reason is added to the comment on `LedgerEntry`.
- `_standings`, `/market leaderboard`, `/market balance`, `/market list` and `market_autocomplete` are all season-agnostic and already filter on `Market.status == "OPEN"`, so carried markets show up with no edits. `/market claim` needs none either — a carrier holds open positions, so the faucet stays shut and selling is the way out.
- A negative cash balance is now possible, though it takes `/market claim` top-ups funding stakes past 10,000 to get there. `/market bet` already gates on the balance, so it just blocks betting until the player sells.
- Parking your whole balance in a market on the last day of the quarter is neutral, not exploitable — the stake is debited straight back.
- The `Clock` in the unit tests returns naive datetimes, which SQLite accepts and Postgres does not — running against pg needs a tz-aware clock to get past `active_season`'s `now() >= season.ends_at`. Production `now()` is always aware, so this is a fidelity gap in the fixture rather than anything this branch changes.

Testing beyond unit tests: ran the merged tree under `docker-compose` against real Postgres 17, since the rollover is all database work and the unit tests only ever see SQLite.

- Startup migrations reach head on an empty database (`001_weather_locations` → `004_quarterly_seasons`), both through `python -m duckbot` and directly.
- A script driving the cog's methods against pg walks the carry path: the market rides into Season 2 still `OPEN` with its position intact, the stake is debited from the new grant (9,500) while a player holding nothing gets a clean 10,000, one `carry` ledger row of -500 is written, `SeasonResult.final_balance` archives on net worth rather than cash, a second boundary charges again, and a later void refunds exactly the 10,000 it was charged.

The clock-advance unit tests drive the same path, including a market carried across two boundaries.

##### Checklist

### DIFF
--- a/duckbot/cogs/playmarket/market.py
+++ b/duckbot/cogs/playmarket/market.py
@@ -342,18 +342,27 @@ class PlayMarket(commands.Cog):
         return season
 
     def _rollover(self, session, season) -> Season:
-        """Void unresolved markets, record final standings, archive the season, and start the next."""
-        for market in session.query(Market).filter(Market.season_id == season.id, Market.status == "OPEN").all():
-            self._resolve_market(session, market, "void")
-        accounts = session.query(PlayerAccount).all()
-        for rank, account in enumerate(sorted(accounts, key=lambda a: a.balance, reverse=True), start=1):
-            session.add(SeasonResult(season_id=season.id, user_id=account.id, final_balance=account.balance, rank=rank))
+        """Record final standings, archive the season, and start the next; open markets ride forward with their positions."""
+        carried = self._carry_costs(session, season)
+        for rank, (user_id, cash, shares_value) in enumerate(self._standings(session), start=1):
+            session.add(SeasonResult(season_id=season.id, user_id=user_id, final_balance=_whole(cash + shares_value), rank=rank))
         season.status = "archived"
         next_season = self._new_season(session, starts_at=season.ends_at)  # keep the calendar alignment
-        for account in accounts:
+        for market in session.query(Market).filter(Market.season_id == season.id, Market.status == "OPEN").all():
+            market.season_id = next_season.id
+        for account in session.query(PlayerAccount).all():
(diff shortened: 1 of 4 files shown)


jameshughes89 (approved):
The rollover choice is a choice

